PR #3, opened 2015-02-18
Modify print_disk_info. Now there is no output if the path is no mountpoint.

chrko:
If there is the tested path is no mountpoint, I do not want any output in the statusbar.
Adding options to configure the format_not_mounted or something else, I do not have the knowlegde at the moment.

### DIFF
--- a/src/print_disk_info.c
+++ b/src/print_disk_info.c
@@ -131,7 +131,7 @@ void print_disk_info(yajl_gen json_gen, char *buffer, const char *path, const ch
                 struct mntent *m;
                 bool found = false;
 
-                while (NULL != (m = getmntent(mntentfile))) {
+                while ((m = getmntent(mntentfile)) != NULL) {
                         if (strcmp(m->mnt_dir, path) == 0) {
                                 found = true;
                                 break;
--- a/src/print_disk_info.c
+++ b/src/print_disk_info.c
@@ -125,7 +125,6 @@ void print_disk_info(yajl_gen json_gen, char *buffer, const char *path, const ch
 
         if (statvfs(path, &buf) == -1)
                 return;
-#endif
 
         FILE *mntentfile = setmntent("/etc/mtab", "r");
         struct mntent *m;
@@ -144,6 +143,7 @@ void print_disk_info(yajl_gen json_gen, char *buffer, const char *path, const ch
                 OUTPUT_FULL_TEXT(buffer);
                 return;
         }
+#endif
 
         if (low_threshold > 0 && below_threshold(buf, prefix_type, threshold_type, low_threshold)) {
                 START_COLOR("color_bad");
--- a/src/print_disk_info.c
+++ b/src/print_disk_info.c
@@ -3,7 +3,9 @@
 #include <stdlib.h>
 #include <string.h>
 #include <ctype.h>
+#include <mntent.h>
 #include <stdint.h>
+#include <sys/stat.h>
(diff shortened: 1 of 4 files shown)


src/print_disk_info.c
```diff
@@ -125,6 +127,24 @@ void print_disk_info(yajl_gen json_gen, char *buffer, const char *path, const ch
                 return;
 #endif
 
+        FILE *mntentfile = setmntent("/etc/mtab", "r");
+        struct mntent *m;
+        bool found = false;
+
+        while ((m = getmntent(mntentfile))) {
```
Merovius: I'd add a != NULL for clarity

Merovius:
A disadvantage of using getmntent etc. instead of a separate statfs call on the parent-directory is that it requires more syscalls (open, multiple reads, close). An advantage is that you don't have to normalize paths or figure out what the "parent" is. Imho it's okay as it is, but you might be able to save some resources if need be.

Also, I would make this optional. Probably by adding a format_unmounted parameter to the module (like the wireless and ethernet modules). That way this change is 100% backwards compatible and people who rather have a custom text for unmounted mountpoints are happy too.

Merovius:
Acoording to [this](https://www.gnu.org/software/gnulib/manual/html_node/setmntent.html) this method might not be available on the BSDs so maybe add some conditional compilation macros.

src/print_disk_info.c
```diff
@@ -125,6 +127,24 @@ void print_disk_info(yajl_gen json_gen, char *buffer, const char *path, const ch
...
+                if (strcmp(m->mnt_dir, path) == 0) {
+                        found = true;
+                        break;
+                }
+        }
+        endmntent(mntentfile);
+
+        if(!found) {
```
Merovius: space after if

src/print_disk_info.c
```diff
@@ -125,6 +127,24 @@ void print_disk_info(yajl_gen json_gen, char *buffer, const char *path, const ch
...
+                *outwalk = '\0';
```
Merovius: For clarity, I would prefer *buffer instead of *outwalk. We don't walk and it makes it clearer that we want to write the empty string.

chrko:
Followed Merovius recommandations in the last 3 commits. Now the documentation needs one additional sentence.

src/print_disk_info.c
```diff
@@ -123,6 +125,24 @@ void print_disk_info(yajl_gen json_gen, char *buffer, const char *path, const ch
 
         if (statvfs(path, &buf) == -1)
                 return;
+
+        if (format_not_mounted != NULL) {
+                FILE *mntentfile = setmntent("/etc/mtab", "r");
+                struct mntent *m;
+                bool found = false;
+
+                while (NULL != (m = getmntent(mntentfile))) {
```
Merovius: Yoda conditions are unnecessary, weird and inconsistent with the rest of the codebase. :)

Merovius:
Add this to the manpage too. Apart from that (and the minor nit above) this LGTM.

stapelberg:
What @Merovius said. Once you’ve added this feature to the manpage and fixed the coding style issue, I’ll merge the pull request.

chrko:
I added some text to the manpage. But I am not sure if we should mention that this new feature is not availabe on BSDs systems, and also the best wording for it.
The example of the feature is in the text enclosed with »«. I think that injecting the sample in the subsequent listing is no solution, because displaying an empty string could be confusing.

Issue #4, opened 2015-02-22
Volume module is very slow on systems with PulseAudio


lahwaacz:
The use of asoundlib in `print_volume.h` results in very poor performance on Linux systems with PulseAudio as the default sound server. I've done a [profiling test](https://github.com/lahwaacz/i3status/commits/profiling) to support my claims, the results are [here](https://github.com/lahwaacz/i3status/blob/profiling/callgrind.out.30029) (viewable with e.g. [kcachegrind](http://kcachegrind.sourceforge.net/html/Home.html)). In summary, during 20 iterations there are 20 calls to `print_volume`, 20 calls to `pa_mainloop_run`, subsequently 3680 calls to `pa_mainloop_iterate`, 120302 calls to `pa_xmalloc` etc.

This may be either a bug in `print_volume.c`, in asoundlib itself or merely a fact showing the need for a native PulseAudio module for i3status -- I don't really know.

For the record, here is my `/etc/asound.conf` file:
```
# Use PulseAudio by default
pcm.!default {
  type pulse
  fallback "sysdefault"
  hint {
    show on
    description "Default ALSA Output (currently PulseAudio Sound Server)"
  }
}

ctl.!default {
  type pulse
  fallback "sysdefault"
}
```

stapelberg:
You’ve stumbled upon this before. I think the library we’re using _shells out_ when you’re using PulseAudio.

If you could send a pull request to address this, that’d be great. Ideally, you’d modify the code in such a way that we don’t need to change the configuration file at all, and it’d just more or less directly talk to PulseAudio when appropriate.

stapelberg:
This has been fixed with #7.

PR #7, opened 2015-03-11
PulseAudio support for volume input

Watcom:
Not sure if this will be useful to anyone, it works for me. If compiled with `make`, nothing is changed (only ALSA is supported). By compiling with `make USE_PULSEAUDIO=1`, PulseAudio support for volume input is available by specifying `device="pulse:N"` line in i3status.conf, where N is the PulseAudio sink index. ALSA is still available if the device line does not match that format.

### DIFF
--- a/i3status.c
+++ b/i3status.c
@@ -62,6 +62,9 @@ cfg_t *cfg, *cfg_general, *cfg_section;
 
 void **cur_instance;
 
+pthread_cond_t i3status_sleep_cond = PTHREAD_COND_INITIALIZER;
+pthread_mutex_t i3status_sleep_mutex = PTHREAD_MUTEX_INITIALIZER;
+
 /*
  * Set the exit_upon_signal flag, because one cannot do anything in a safe
  * manner in a signal handler (e.g. fprintf, which we really want to do for
@@ -547,6 +550,7 @@ int main(int argc, char *argv[]) {
     char buffer[4096];
 
     void **per_instance = calloc(cfg_size(cfg, "order"), sizeof(*per_instance));
+    pthread_mutex_lock(&i3status_sleep_mutex);
 
     while (1) {
         if (exit_upon_signal) {
@@ -682,13 +686,16 @@ int main(int argc, char *argv[]) {
         fflush(stdout);
 
         /* To provide updates on every full second (as good as possible)
-         * we don’t use sleep(interval) but we sleep until the next
-         * second (with microsecond precision) plus (interval-1)
-         * seconds. We also align to 60 seconds modulo interval such
+         * we don’t use sleep(interval) but we sleep until the next second.
+         * We also align to 60 seconds modulo interval such
          * that we start with :00 on every new minute. */
-        struct timeval current_timeval;
-        gettimeofday(&current_timeval, NULL);
-        struct timespec ts = {interval - 1 - (current_timeval.tv_sec % interval), (10e5 - current_timeval.tv_usec) * 1000};
-        nanosleep(&ts, NULL);
+        struct timespec ts;
+        clock_gettime(CLOCK_REALTIME, &ts);
+        ts.tv_sec += interval - (ts.tv_sec % interval);
+        ts.tv_nsec = 0;
+
+        /* Sleep to absolute time 'ts', unless the condition
+         * 'i3status_sleep_cond' is signaled from another thread */
+        pthread_cond_timedwait(&i3status_sleep_cond, &i3status_sleep_mutex, &ts);
     }
 }
--- a/include/i3status.h
+++ b/include/i3status.h
@@ -14,9 +14,13 @@ enum { O_DZEN2,
 #include <yajl/yajl_version.h>
 #include <unistd.h>
 #include <string.h>
+#include <pthread.h>
+#include <stdint.h>
 
 #define BEGINS_WITH(haystack, needle) (strncmp(haystack, needle, strlen(needle)) == 0)
 #define max(a, b) ((a) > (b) ? (a) : (b))
+
+#define DEFAULT_SINK_INDEX UINT32_MAX
 
 #if defined(LINUX)
 
@@ -195,6 +199,8 @@ void print_eth_info(yajl_gen json_gen, char *buffer, const char *interface, cons
 void print_load(yajl_gen json_gen, char *buffer, const char *format, const float max_threshold);
 void print_volume(yajl_gen json_gen, char *buffer, const char *fmt, const char *fmt_muted, const char *device, const char *mixer, int mixer_idx);
 bool process_runs(const char *path);
+int volume_pulseaudio(uint32_t sink_idx);
+bool pulse_initialize(void);
 
 /* socket file descriptor for general purposes */
 extern int general_socket;
@@ -202,5 +208,8 @@ extern int general_socket;
 extern cfg_t *cfg, *cfg_general, *cfg_section;
 
 extern void **cur_instance;
+
+extern pthread_cond_t i3status_sleep_cond;
+extern pthread_mutex_t i3status_sleep_mutex;
 
 #endif
--- a/src/print_volume.c
+++ b/src/print_volume.c
@@ -51,13 +51,40 @@ void print_volume(yajl_gen json_gen, char *buffer, const char *fmt, const char *
     char *outwalk = buffer;
     int pbval = 1;
 
-    /* Printing volume only works with ALSA at the moment */
+    /* Printing volume works with ALSA and PulseAudio at the moment */
     if (output_format == O_I3BAR) {
         char *instance;
         asprintf(&instance, "%s.%s.%d", device, mixer, mixer_idx);
         INSTANCE(instance);
         free(instance);
     }
+
+    /* Try PulseAudio first */
+
+    /* If the device name has the format "pulse[:N]" where N is the
+     * index of the PulseAudio sink then force PulseAudio, optionally
+     * overriding the default sink */
+    if (!strncasecmp(device, "pulse", strlen("pulse"))) {
+        uint32_t sink_idx = device[5] == ':' ? (uint32_t)atoi(device + 6)
+                                             : DEFAULT_SINK_INDEX;
+        int ivolume = pulse_initialize() ? volume_pulseaudio(sink_idx) : 0;
+        /* negative result means error, stick to 0 */
+        if (ivolume < 0)
+            ivolume = 0;
+        outwalk = apply_volume_format(fmt, outwalk, ivolume);
+        goto out;
+    } else if (!strcasecmp(device, "default") && pulse_initialize()) {
+        /* no device specified or "default" set */
+        int ivolume = volume_pulseaudio(DEFAULT_SINK_INDEX);
+        if (ivolume >= 0) {
+            outwalk = apply_volume_format(fmt, outwalk, ivolume);
+            goto out;
+        }
+        /* negative result means error, fail PulseAudio attempt */
+    }
+/* If some other device was specified or PulseAudio is not detected,
+ * proceed to ALSA / OSS */
+
 #ifdef LINUX
     int err;
     snd_mixer_t *m;
--- a/src/pulse.c
+++ b/src/pulse.c
@@ -1,0 +1,245 @@
+// vim:ts=4:sw=4:expandtab
+#include <string.h>
+#include <stdio.h>
+#include <pulse/pulseaudio.h>
+#include "i3status.h"
+#include "queue.h"
+
+#define APP_NAME "i3status"
+#define APP_ID "org.i3wm"
+
+typedef struct indexed_volume_s {
+    uint32_t idx;
+    int volume;
+    TAILQ_ENTRY(indexed_volume_s) entries;
+} indexed_volume_t;
+
+static pa_threaded_mainloop *main_loop = NULL;
+static pa_context *context = NULL;
+static pa_mainloop_api *api = NULL;
+static bool context_ready = false;
+static uint32_t default_sink_idx = DEFAULT_SINK_INDEX;
+TAILQ_HEAD(tailhead, indexed_volume_s) cached_volume =
+    TAILQ_HEAD_INITIALIZER(cached_volume);
+static pthread_mutex_t pulse_mutex = PTHREAD_MUTEX_INITIALIZER;
+
+static void pulseaudio_error_log(pa_context *c) {
+    fprintf(stderr,
+            "i3status: PulseAudio: %s\n",
+            pa_strerror(pa_context_errno(c)));
+}
+
+static bool pulseaudio_free_operation(pa_context *c, pa_operation *o) {
+    if (o)
+        pa_operation_unref(o);
+    else
+        pulseaudio_error_log(c);
+    /* return false if the operation failed */
+    return o;
+}
+
+/*
+ * save the volume for the specified sink index
+ * returning true if the value was changed
+ */
+static bool save_volume(uint32_t sink_idx, int new_volume) {
+    pthread_mutex_lock(&pulse_mutex);
+    indexed_volume_t *entry;
+    TAILQ_FOREACH(entry, &cached_volume, entries) {
+        if (entry->idx == sink_idx) {
+            const bool changed = (new_volume != entry->volume);
+            entry->volume = new_volume;
+            pthread_mutex_unlock(&pulse_mutex);
+            return changed;
+        }
+    }
+    /* index not found, store it */
+    entry = malloc(sizeof(*entry));
+    TAILQ_INSERT_HEAD(&cached_volume, entry, entries);
+    entry->idx = sink_idx;
+    entry->volume = new_volume;
+    pthread_mutex_unlock(&pulse_mutex);
+    return true;
+}
+
+static void store_volume_from_sink_cb(pa_context *c,
+                                      const pa_sink_info *info,
+                                      int eol,
+                                      void *userdata) {
+    if (eol < 0) {
+        if (pa_context_errno(c) == PA_ERR_NOENTITY)
+            return;
+
+        pulseaudio_error_log(c);
+        return;
+    }
+
+    if (eol > 0)
+        return;
+
+    int avg_vol = pa_cvolume_avg(&info->volume);
+    int vol_perc = (int)((long long)avg_vol * 100 / PA_VOLUME_NORM);
+
+    /* if this is the default sink we must try to save it twice: once with
+     * DEFAULT_SINK_INDEX as the index, and another with its proper value
+     * (using bitwise OR to avoid early-out logic) */
+    if ((info->index == default_sink_idx &&
+         save_volume(DEFAULT_SINK_INDEX, vol_perc)) |
+        save_volume(info->index, vol_perc)) {
+        /* if the volume changed, wake the main thread */
+        pthread_mutex_lock(&i3status_sleep_mutex);
+        pthread_cond_broadcast(&i3status_sleep_cond);
+        pthread_mutex_unlock(&i3status_sleep_mutex);
+    }
+}
+
+static void get_sink_info(pa_context *c, uint32_t idx) {
+    pa_operation *o =
+        idx == DEFAULT_SINK_INDEX ? pa_context_get_sink_info_by_name(
+                                        c, "@DEFAULT_SINK@", store_volume_from_sink_cb, NULL)
+                                  : pa_context_get_sink_info_by_index(
+                                        c, idx, store_volume_from_sink_cb, NULL);
+    pulseaudio_free_operation(c, o);
+}
+
+static void store_default_sink_cb(pa_context *c,
+                                  const pa_sink_info *i,
+                                  int eol,
+                                  void *userdata) {
+    if (i) {
+        if (default_sink_idx != i->index) {
+            /* default sink changed? */
+            default_sink_idx = i->index;
+            store_volume_from_sink_cb(c, i, eol, userdata);
+        }
+    }
+}
+
+static void update_default_sink(pa_context *c) {
+    pa_operation *o = pa_context_get_sink_info_by_name(
+        c,
+        "@DEFAULT_SINK@",
+        store_default_sink_cb,
+        NULL);
+    pulseaudio_free_operation(c, o);
+}
+
+static void subscribe_cb(pa_context *c, pa_subscription_event_type_t t,
+                         uint32_t idx, void *userdata) {
+    if ((t & PA_SUBSCRIPTION_EVENT_TYPE_MASK) != PA_SUBSCRIPTION_EVENT_CHANGE)
+        return;
+    pa_subscription_event_type_t facility =
+        t & PA_SUBSCRIPTION_EVENT_FACILITY_MASK;
+    switch (facility) {
+        case PA_SUBSCRIPTION_EVENT_SERVER:
+            /* server change event, see if the default sink changed */
+            update_default_sink(c);
+            break;
+        case PA_SUBSCRIPTION_EVENT_SINK:
+            get_sink_info(c, idx);
+            break;
+        default:
+            break;
+    }
+}
+
+static void context_state_callback(pa_context *c, void *userdata) {
+    switch (pa_context_get_state(c)) {
+        case PA_CONTEXT_UNCONNECTED:
+        case PA_CONTEXT_CONNECTING:
+        case PA_CONTEXT_AUTHORIZING:
+        case PA_CONTEXT_SETTING_NAME:
+        case PA_CONTEXT_TERMINATED:
+        default:
+            break;
+
+        case PA_CONTEXT_READY: {
+            pa_context_set_subscribe_callback(c, subscribe_cb, NULL);
+            update_default_sink(c);
+
+            pa_operation *o = pa_context_subscribe(
+                c,
+                PA_SUBSCRIPTION_MASK_SINK | PA_SUBSCRIPTION_MASK_SERVER,
+                NULL,
+                NULL);
+            if (!pulseaudio_free_operation(c, o))
+                break;
+            context_ready = true;
+        } break;
+
+        case PA_CONTEXT_FAILED:
+            pulseaudio_error_log(c);
+            break;
+    }
+}
+
+/*
+ * returns the current volume in percent, which, as per PulseAudio,
+ * may be > 100%
+ */
+int volume_pulseaudio(uint32_t sink_idx) {
+    if (!context_ready || default_sink_idx == DEFAULT_SINK_INDEX)
+        return -1;
+
+    pthread_mutex_lock(&pulse_mutex);
+    const indexed_volume_t *entry;
+    TAILQ_FOREACH(entry, &cached_volume, entries) {
+        if (entry->idx == sink_idx) {
+            int vol = entry->volume;
+            pthread_mutex_unlock(&pulse_mutex);
+            return vol;
+        }
+    }
+    pthread_mutex_unlock(&pulse_mutex);
+    /* first time requires a prime callback call because we only get
+     * updates when the volume actually changes, but we need it to
+     * be correct even if it never changes */
+    pa_threaded_mainloop_lock(main_loop);
+    get_sink_info(context, sink_idx);
+    pa_threaded_mainloop_unlock(main_loop);
+    /* show 0 while we don't have this information */
+    return 0;
+}
+
+/*
+ *  detect and, if necessary, initialize the PulseAudio API
+ */
+bool pulse_initialize(void) {
+    if (!main_loop) {
+        main_loop = pa_threaded_mainloop_new();
+        if (!main_loop)
+            return false;
+    }
+    if (!api) {
+        api = pa_threaded_mainloop_get_api(main_loop);
+        if (!api)
+            return false;
+    }
+    if (!context) {
+        pa_proplist *proplist = pa_proplist_new();
+        pa_proplist_sets(proplist, PA_PROP_APPLICATION_NAME, APP_NAME);
+        pa_proplist_sets(proplist, PA_PROP_APPLICATION_ID, APP_ID);
+        pa_proplist_sets(proplist, PA_PROP_APPLICATION_VERSION, VERSION);
+        context = pa_context_new_with_proplist(api, APP_NAME, proplist);
+        pa_proplist_free(proplist);
+        if (!context)
+            return false;
+        pa_context_set_state_callback(context,
+                                      context_state_callback,
+                                      NULL);
+        if (pa_context_connect(context,
+                               NULL,
+                               PA_CONTEXT_NOFAIL | PA_CONTEXT_NOAUTOSPAWN,
+                               NULL) < 0) {
+            pulseaudio_error_log(context);
+            return false;
+        }
+        if (pa_threaded_mainloop_start(main_loop) < 0) {
+            pulseaudio_error_log(context);
+            pa_threaded_mainloop_free(main_loop);
+            main_loop = NULL;
+            return false;
+        }
+    }
+    return true;
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ before_install:
   - sudo apt-get install -t utopic clang-format-3.5
   - clang-format-3.5 --version
 install:
-  - sudo apt-get install libconfuse-dev libyajl-dev libasound2-dev libiw-dev asciidoc libcap2-bin
+  - sudo apt-get install libconfuse-dev libyajl-dev libasound2-dev libiw-dev asciidoc libcap2-bin libpulse-dev
 script:
   - make -j
   - clang-format-3.5 -i **/*.[ch] && git diff --exit-code || (echo 'Code was not formatted using clang-format!'; false)
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,7 @@ CPPFLAGS+=-DVERSION=\"${GIT_VERSION}\"
 CFLAGS+=-Iinclude
 LIBS+=-lconfuse
 LIBS+=-lyajl
+LIBS+=-lpulse
 
 VERSION:=$(shell git describe --tags --abbrev=0)
 GIT_VERSION:="$(shell git describe --tags --always) ($(shell git log --pretty=format:%cd --date=short -n1))"
--- a/README
+++ b/README
@@ -20,9 +20,10 @@ i3status has the following dependencies:
  • libiw-dev
  • libcap2-bin (for getting network status without root permissions)
  • asciidoc (only for the documentation)
+ • libpulse-dev (for getting the current volume using PulseAudio)
 
 On debian-based systems, the following line will install all requirements:
-apt-get install libconfuse-dev libyajl-dev libasound2-dev libiw-dev asciidoc libcap2-bin
+apt-get install libconfuse-dev libyajl-dev libasound2-dev libiw-dev asciidoc libcap2-bin libpulse-dev
 
  ┌────────────────────────────┐
  │ Upstream                   │
--- a/man/i3status.man
+++ b/man/i3status.man
@@ -422,13 +422,26 @@ details on the format string.
 
 === Volume
 
-Outputs the volume of the specified mixer on the specified device. Works only
-on Linux because it uses ALSA.
-A simplified configuration can be used on FreeBSD and OpenBSD due to
-the lack of ALSA,  the +device+ and +mixer+ options can be
-ignored on these systems. On these systems the OSS API is used instead to
-query +/dev/mixer+ directly if +mixer_dix+ is -1, otherwise
-+/dev/mixer++mixer_idx+.
+Outputs the volume of the specified mixer on the specified device.  PulseAudio
+and ALSA (Linux only) are supported.  If PulseAudio is absent, a simplified
+configuration can be used on FreeBSD and OpenBSD due to the lack of ALSA,  the
++device+ and +mixer+ options can be ignored on these systems. On these systems
+the OSS API is used instead to query +/dev/mixer+ directly if +mixer_idx+ is
+-1, otherwise +/dev/mixer++mixer_idx+.
+
+To get PulseAudio volume information, one must use the following format in the
+device line:
+
+ device = "pulse"
+
+or
+
+ device = "pulse:N"
+
+where N is the index of the PulseAudio sink. If no sink is specified the
+default is used. If the device string is missing or is set to "default",
+PulseAudio will be tried if detected and will fallback to ALSA (Linux)
+or OSS (FreeBSD/OpenBSD).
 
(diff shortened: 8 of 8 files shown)


stapelberg:
Before we dive into any of the details of the pull request itself, can we make it so that PulseAudio is automatically queried when it’s in use on the user’s system? The current ALSA implementation faills back to PulseAudio, but rather poorly, see https://github.com/i3/i3status/issues/4

Watcom:
You mean if the user does not specify a device? Now that you mention it, it makes sense to try the default sink in this case and, in case of failure, fall back to ALSA. I'll try that.

Watcom:
Ok, now PulseAudio is tried first, then ALSA. Also, now the default sink is fetched and used if not explicitly specified. If the device specified is anything other than "pulse" or "default" (or missing), then PulseAudio is not used even if it's available.

stapelberg:
Thanks for making that change.

I compiled and ran this, but it seems like even when using libpulse, there still is a fork() involved every second in order to get the volume:

```
$ strace -tt -f -o /tmp/st -s 2048 ./i3status
^C
$ grep clone /tmp/st
9699  09:42:20.121223 clone(child_stack=0x7fc3fc864f30, flags=CLONE_VM|CLONE_FS|CLONE_FILES|CLONE_SIGHAND|CLONE_THREAD|CLONE_SYSVSEM|CLONE_SETTLS|CLONE_PARENT_SETTID|CLONE_CHILD_CLEARTID, parent_tidptr=0x7fc3fc8659d0, tls=0x7fc3fc865700, child_tidptr=0x7fc3fc8659d0) = 9700
9699  09:42:21.008604 clone(child_stack=0x7fc3fc864f30, flags=CLONE_VM|CLONE_FS|CLONE_FILES|CLONE_SIGHAND|CLONE_THREAD|CLONE_SYSVSEM|CLONE_SETTLS|CLONE_PARENT_SETTID|CLONE_CHILD_CLEARTID, parent_tidptr=0x7fc3fc8659d0, tls=0x7fc3fc865700, child_tidptr=0x7fc3fc8659d0) = 9702
9699  09:42:22.007302 clone(child_stack=0x7fc3fc864f30, flags=CLONE_VM|CLONE_FS|CLONE_FILES|CLONE_SIGHAND|CLONE_THREAD|CLONE_SYSVSEM|CLONE_SETTLS|CLONE_PARENT_SETTID|CLONE_CHILD_CLEARTID, parent_tidptr=0x7fc3fc8659d0, tls=0x7fc3fc865700, child_tidptr=0x7fc3fc8659d0) = 9704
```

Is that because of how this code uses the library or is it inherent to how the library works?

Also I’m wondering how libasound can avoid being linked against all these dependencies that libpulse brings in but still fall back to query pulseaudio for volume.

An alternative could be using dbus, but it seems like PulseAudio is shipped with the dbus module disabled on plenty of linux distributions :(.

I haven’t looked into it in detail, but would it be feasible to directly talk to PulseAudio without using libpulse?

Watcom:
That's odd. I tested that here and I see no forking. Perhaps it's falling back to ALSA for some reason in your system?

How is your device= line in i3status.conf? Make sure it's either unset or set to "default" or "pulse". One quick way to see if it's using PulseAudio is by setting the sink volume to above 100% (e.g. with pavucontrol). It will only show values above 100% if using PulseAudio, but not with ALSA.

Watcom:
Should I change the default to building with PulseAudio support and, subsequently, the make flag from `USE_PULSEAUDIO=1` to `NO_PULSEAUDIO=1`?

stapelberg:
The system in question is indeed using PulseAudio. After repeating and making sure I compile with `make USE_PULSEAUDIO=1`, I cannot reproduce the forking, so I think I might have screwed up in my initial test.

I’ve done some digging in the PulseAudio API, and have two high-level comments:

1. You can use the `@DEFAULT_SINK@` sink name to avoid one level of callbacks. In fact, you don’t need the server_info callback either, I think, so you can get by with just one callback.

2. Instead of using the simple mainloop and having the volume lag behind one iteration (unclean), you could use the `pa_threaded_mainloop` much like the simple API is using it. I’ve written a little test program to illustrate what I mean: https://gist.github.com/stapelberg/507b400483e2f8efdebd

With regards to building: please just remove the conditionals entirely and make libpulse a regular dependency. These days PulseAudio is wide-spread, and having it optional will be more of a headache than it buys us.

By the way, I think it’d be worth a shot to add a function to get the volume (or the sink info in general) to libpulse’s simple API. That way, we could eventually switch to just using their simple API, as we are basically replicating it (after going the route I suggested). Is that something you’d want to do?

Watcom:
> You can use the @DEFAULT_SINK@ sink name to avoid one level of callbacks. In fact, you don’t need the server_info callback either, I think, so you can get by with just one callback.

Nice catch, I completely missed that one. Done, and indeed, got rid of two callbacks.

> With regards to building: please just remove the conditionals entirely and make libpulse a regular dependency.

Done. The added dependency seems to be failing the automated builds, though.

> Also, can you rebase your commits on top of master and squash your history? I.e., I’m not particularly interested in your “clang-format” commit :).

Sorry about that, it's squashed now.

> Instead of using the simple mainloop and having the volume lag behind one iteration (unclean), you could use the pa_threaded_mainloop much like the simple API is using it.

Originally I considered using pa_threaded_mainloop but I had doubts about the proper way to use it and decided to go the safe way. In particular, I thought of subscribing to volume changes from the sink and waking the main thread upon volume changes so that a refresh was issued when required. I don't know if this behaviour is always desirable, hence my doubt (maybe something configurable?). I can try this direction if you think it's ok.

> By the way, I think it’d be worth a shot to add a function to get the volume (or the sink info in general) to libpulse’s simple API. That way, we could eventually switch to just using their simple API, as we are basically replicating it (after going the route I suggested). Is that something you’d want to do?

I think if we subscribe to volume changes like I've suggested then the simple API could not be used.

stapelberg:
> Originally I considered using pa_threaded_mainloop but I had doubts about the proper way to use it and decided to go the safe way. In particular, I thought of subscribing to volume changes from the sink and waking the main thread upon volume changes so that a refresh was issued when required. I don't know if this behaviour is always desirable, hence my doubt (maybe something configurable?). I can try this direction if you think it's ok.

Ah, I didn’t see that one can subscribe to volume changes. That’d be the most elegant solution, I think (subscribing, modifying a variable when the volume changes, and just returning that variable in each loop iteration).

I don’t think we should make this configurable, because why would you not want it to work that way?

Watcom:
> I don’t think we should make this configurable, because why would you not want it to work that way?

I meant waking the main thread so that i3status would print the new volume immediately.

Currently I have something like this in my .i3/config for changing the PulseAudio volume from the keyboard:

```
bindsym XF86AudioRaiseVolume exec --no-startup-id pactl set-sink-volume 1 -- +2% && killall -SIGUSR1 i3status
bindsym XF86AudioLowerVolume exec --no-startup-id pactl set-sink-volume 1 -- -2% && killall -SIGUSR1 i3status
```

The `killall -SIGUSR1 i3status` would not be needed if i3status knew the volume changed and refreshed immediately.

stapelberg:
Even tweaking the main thread to wake up is something I’d be willing to do without a configuration option, as long as the volume change event doesn’t trigger when it shouldn’t.

Watcom:
Ok, converted it to a threaded mainloop plus waking of the main thread when the volume changes. Seems to be working quite well.

src/print_volume.c
```diff
@@ -24,20 +24,79 @@
...
         free(instance);
     }
+
+    /* Try PulseAudio first */
+
+    /* If the device name has the format "pulse[:N]" where N is the
+         * index of the PulseAudio sink then force PulseAudio, optionally
+         * overriding the default sink */
+
+    if (device[0] == 'p' &&
```
stapelberg: Use `strncasecmp(device, "pulse", strlen("pulse"))` instead.

src/print_volume.c
```diff
@@ -24,20 +24,79 @@
...
+        goto out;
+    } else if (!strcmp(device, "default") && pulse_detected()) {
+        /* no device specified or "default" set */
+        int ivolume = volume_pulseaudio(-1);
+        if (ivolume >= 0) {
+            outwalk = apply_volume_format(fmt, outwalk, ivolume);
+            goto out;
+        } /* negative result means error, fail PulseAudio attempt */
+    }
+/* If some other device was specified or PulseAudio is not detected,
```
stapelberg: Comment formatting seems broken.

Also please remove the commented-out fprintf. Either it should be in the code or it shouldn’t, there is no in-between :).

src/print_volume.c
```diff
@@ -113,21 +172,8 @@ void print_volume(yajl_gen json_gen, char *buffer, const char *fmt, const char *
...
-        if (BEGINS_WITH(walk + 1, "%")) {
-            outwalk += sprintf(outwalk, "%%");
-            walk += strlen("%");
-        }
-        if (BEGINS_WITH(walk + 1, "volume")) {
-            outwalk += sprintf(outwalk, "%d%%", avg);
-            walk += strlen("volume");
-        }
-    }
+    outwalk = apply_volume_format(fmt, outwalk, avg);
```
stapelberg: The refactoring here is unrelated to the introduction of pulseaudio, isn’t it? If so, can you please move that into a separate commit (or even pull request)?

src/pulse.c
```diff
@@ -0,0 +1,230 @@
...
+#include <string.h>
+#include <stdio.h>
+#include <pulse/pulseaudio.h>
+#include "pulse.h"
+#include "i3status.h"
+
+#define APP_NAME "i3status"
+#define APP_ID "org.i3wm"
+#define stringize(x) #x
+#define APP_VERSION stringize(VERSION)
```
stapelberg: Use `#define APP_VERSION #VERSION`, since you don’t use `stringize` anywhere else.

src/pulse.c
```diff
@@ -0,0 +1,230 @@
...
+
+static void sink_cb(pa_context *c,
+                    const pa_sink_info *info,
+                    int eol,
+                    void *userdata) {
+    if (eol < 0) {
+        if (pa_context_errno(context) == PA_ERR_NOENTITY)
+            return;
+
+        fprintf(stderr, "i3status: sink_cb: link callback failure\n");
```
stapelberg: s/link/sink/. Can we get a better error message here, i.e. does pulseaudio offer us an `strerror()`-like function?

src/pulse.c
```diff
@@ -0,0 +1,230 @@
...
+
+    if (eol > 0)
+        return;
+
+    int avg_vol = pa_cvolume_avg(&info->volume);
+    int vol_perc = (int)((long long)avg_vol * 100 / PA_VOLUME_NORM);
+    volatile int *pvolume = volume_slot_from_sink_idx(info->index);
+    int old_value = *pvolume;
+
+    if (old_value != vol_perc) {
```
stapelberg: Use this idiom to get rid of one level of indentation:

```c
if (old_value == vol_perc)
    return;
```

src/pulse.c
```diff
@@ -0,0 +1,230 @@
...
+    }
+}
+
+/*
+ * subscribe_cb
+ */
+
+static void subscribe_cb(pa_context *c, pa_subscription_event_type_t t,
+                         uint32_t index, void *userdata) {
+    switch (t & PA_SUBSCRIPTION_EVENT_FACILITY_MASK) {
```
stapelberg: This seems long-winded. I suggest:

```c
if ((t & PA_SUBSCRIPTION_EVENT_FACILITY_MASK) != PA_SUBSCRIPTION_EVENT_SINK)
    return;

if ((t & PA_SUBSCRIPTION_EVENT_TYPE_MASK) != PA_SUBSCRIPTION_EVENT_CHANGE)
    return;
```

src/pulse.c
```diff
@@ -0,0 +1,230 @@
...
+        case PA_SUBSCRIPTION_EVENT_SINK:
+            if ((t & PA_SUBSCRIPTION_EVENT_TYPE_MASK) ==
+                PA_SUBSCRIPTION_EVENT_CHANGE) {
+                pa_operation *o;
+
+                if (!(o = pa_context_get_sink_info_by_index(
+                          c, index, sink_cb, NULL))) {
+                    fprintf(stderr, "i3status: error getting sink info\n");
```
stapelberg: strerror (see above)?

src/pulse.c
```diff
@@ -0,0 +1,230 @@
...
+            pa_context_set_subscribe_callback(c,
+                                              subscribe_cb,
+                                              NULL);
+
+            if (!(o = pa_context_subscribe(
+                      c,
+                      (pa_subscription_mask_t)(PA_SUBSCRIPTION_MASK_SINK),
+                      NULL,
+                      NULL))) {
+                fprintf(stderr, "i3status: pa_context_subscribe() failed");
```
stapelberg: strerror (see above)?

src/pulse.c
```diff
@@ -0,0 +1,230 @@
...
+    return vol;
+}
+
+/*
+ *  pulse_detected
+ *
+ *  detect and, if necessary, initialize the PulseAudio API
+ */
+
+int pulse_detected(void) {
```
stapelberg: I suggest renaming this to `pulse_initialize` since that’s the bulk of what it does.

include/pulse.h
```diff
@@ -0,0 +1,9 @@
+/* pulse.h */
```
stapelberg: Why did you put these forward definitions into their own header file instead of just into `i3status.h` where all the others are? That way, you’d also avoid the Makefile change entirely.

i3status.c
```diff
@@ -62,6 +62,9 @@ cfg_t *cfg, *cfg_general, *cfg_section;
 
 void **cur_instance;
 
+pthread_cond_t i3status_sleep_cond;
```
stapelberg: Use this to avoid the initialization later:

```c
pthread_cond_t i3status_sleep_cond = PTHREAD_COND_INITIALIZER;
pthread_mutex_t i3status_sleep_mutex = PTHREAD_MUTEX_INITIALIZER;
```

src/pulse.c
```diff
@@ -0,0 +1,230 @@
...
+/*
+ * sink_info_callback
+ */
+
+static void sink_info_callback(pa_context *c,
+                               const pa_sink_info *i,
+                               int eol,
+                               void *userdata) {
+    if (i)
+        default_sink_idx = i->index; /* save the index of the default sink */
```
stapelberg: Hm, can the default sink not change during runtime? We’d still have the old value stored in that case.

src/pulse.c
```diff
@@ -0,0 +1,230 @@
...
+ */
+
+int volume_pulseaudio(int sink_idx) {
+    if (!main_loop || !context)
+        return -1;
+
+    int vol = *volume_slot_from_sink_idx(sink_idx < 0 ? 0 : sink_idx);
+
+    if (vol == -2) {
+        /* first time requires prime callback call */
```
stapelberg: Is it necessary to call `pa_context_get_sink_info_by_*` both here and in `context_state_callback`?

src/pulse.c
```diff
@@ -0,0 +1,230 @@
...
+#define APP_VERSION stringize(VERSION)
+
+#define MAX_SINKS 16 /* must be a power of 2 */
+
+static pa_threaded_mainloop *main_loop = NULL;
+static pa_context *context = NULL;
+static pa_mainloop_api *api = NULL;
+static int initialized = 0;
+static int default_sink_idx = -1;
+static volatile int cached_volume[MAX_SINKS];
```
stapelberg: I dislike that this is static and limited to an arbitrary number of 16. Can you make it dynamic so that we never run into the problem of too many sinks being present on a system?

src/print_volume.c
Watcom: Done.

src/print_volume.c
Watcom: Done.

src/print_volume.c
Watcom: Made a separate pull request.

src/pulse.c
Watcom: I just realized my code is actually wrong here since it would expand to literally "#VERSION", whereas I want to expand the VERSION macro first. I had to add yet another macro level, this is how silly the C preprocessor is. Your suggestion wouldn't work because the stringizing operator only works on macro arguments. See a test I made here https://gist.github.com/Watcom/90ff692df5c60ba05bbe

src/pulse.c
Watcom: Done.

src/pulse.c
Watcom: Done.

src/pulse.c
Watcom: Done.

src/pulse.c
Watcom: Done.

src/pulse.c
Watcom: Done.

src/pulse.c
Watcom: Done.

include/pulse.h
Watcom: Done, incorporated to `i3status.h`. I originally thought of isolating the PulseAudio-related code as much as possible and making it optional but since you intend to make it a permanent dependency it no longer makes sense.

i3status.c
Watcom: Done.

src/pulse.c
Watcom: Probably, since you can (un)plug devices and PulseAudio is OK with that. I will give it some thought to find a better solution to this problem.

src/pulse.c
Watcom: We'd need a way to store key/value pairs efficiently (indices can have any positive value in PulseAudio). I'm open to suggestions here, as this can potentially add some bloat.

stapelberg: I think just having a TAILQ (from include/queue.h) with a struct that contains key/value is good enough. With the small number of sinks that most systems have, the list traversal shouldn’t be too hard on performance. In case it turns out it actually is benefitial to use some sort of hash table, we can always revisit that later.

src/pulse.c
```diff
@@ -0,0 +1,234 @@
...
+#include <string.h>
+#include <stdio.h>
+#include <pulse/pulseaudio.h>
+#include "i3status.h"
+
+#define APP_NAME "i3status"
+#define APP_ID "org.i3wm"
+#define stringize(x) #x
+#define stringize_content(x) stringize(x)
+#define APP_VERSION stringize(VERSION)
```
stapelberg: Why do you need to stringize this in the first place? In the Makefile, we set `-DVERSION=\"${GIT_VERSION}\"`, and in fact we do use that directly as a string in the `printf` calls in i3status.c.

Watcom: Oh sorry, I missed the escaped quotes the first time I saw it.

src/pulse.c
Watcom: Ok, I will have a look at TAILQ.

Watcom:
The default sink is now kept updated by using a server change subscription. I tested here flipping the default sink between two devices and it updates immediately, alternating the display of each of the device's current volume.

The static array for volume storage was replaced by a dynamically allocated array of struct. There's no longer a hard limit on the number of sinks supported.

src/pulse.c
```diff
@@ -0,0 +1,319 @@
...
+static void store_wanted_sink(int sink_idx) {
+    pthread_mutex_lock(&pulse_mutex);
+    if (main_loop && context)
+        pa_threaded_mainloop_lock(main_loop);
+
+    int idx = find_sink_idx(sink_idx);
+
+    if (idx < 0) /* we haven't seen this index before, store it */
+    {
+        if (used_cached_volume_elements * sizeof(*cached_volume) >=
```
stapelberg: Instead of doing all this reallocation yourself, why not use the `TAILQ` macros I suggested?

src/pulse.c
```diff
@@ -0,0 +1,319 @@
...
+ */
+
+static void store_wanted_sink(int sink_idx) {
+    pthread_mutex_lock(&pulse_mutex);
+    if (main_loop && context)
+        pa_threaded_mainloop_lock(main_loop);
+
+    int idx = find_sink_idx(sink_idx);
+
+    if (idx < 0) /* we haven't seen this index before, store it */
```
stapelberg: nit: can you move all commants that are at the end of the line to the next line? That’s more consistent with our code base.

src/pulse.c
```diff
@@ -0,0 +1,319 @@
...
+    return -1; /* not found */
+}
+
+/*
+ * store_wanted_sink
+ *
+ * store the desired sink index so that the callback can know we're
+ * interested in its volume -- an index of -1 means the @DEFAULT_SINK@
+ */
+
```
stapelberg: nit: here and elsewhere, please kill empty lines between comment and function.

src/pulse.c
```diff
@@ -0,0 +1,319 @@
...
+    }
+    if (!context) {
+        pa_proplist *proplist = pa_proplist_new();
+        pa_proplist_sets(proplist, PA_PROP_APPLICATION_NAME, APP_NAME);
+        pa_proplist_sets(proplist, PA_PROP_APPLICATION_ID, APP_ID);
+        pa_proplist_sets(proplist, PA_PROP_APPLICATION_VERSION, VERSION);
+        context = pa_context_new_with_proplist(api, APP_NAME, proplist);
+        if (!context)
+            return 0;
+        pa_proplist_free(proplist);
```
stapelberg: I think the `pa_proplist_free()` call should also be executed if context cannot be allocated, right?

src/pulse.c
```diff
@@ -0,0 +1,319 @@
...
+            pa_operation *o;
+
+            pa_context_set_subscribe_callback(c,
+                                              subscribe_cb,
+                                              NULL);
+            update_default_sink(c);
+
+            if (!(o = pa_context_subscribe(
+                      c,
+                      (pa_subscription_mask_t)(PA_SUBSCRIPTION_MASK_SINK |
```
stapelberg: Is the typecast to `pa_subscription_mask_t` necessary? If so, why?

src/pulse.c
```diff
@@ -0,0 +1,319 @@
...
+    if (eol > 0)
+        return;
+
+    int avg_vol = pa_cvolume_avg(&info->volume);
+    int vol_perc = (int)((long long)avg_vol * 100 / PA_VOLUME_NORM);
+
+    /* if this is the default sink we must try to save it twice: once
+     * with -1 as the index (meaning default sink), and another with
+     * its proper value (using bitwise OR to avoid early-out logic) */
+    if ((info->index == (unsigned)default_sink_idx &&
```
stapelberg: nit: here and elsewhere, please use explicit data types such as `uint32_t` (which the pulseaudio API uses) instead of just “unsigned”.

Could we avoid the typecasting entirely by making the functions that return a cached volume allocate that entry if necessary instead of returning -1 for not found?

src/pulse.c
```diff
@@ -0,0 +1,319 @@
...
+}
+
+static void get_sink_info(pa_context *c, int idx) {
+    pa_operation *o = idx < 0 ? pa_context_get_sink_info_by_name(
+                                    c, "@DEFAULT_SINK@", store_volume_from_sink_cb, NULL)
+                              : pa_context_get_sink_info_by_index(
+                                    c, (uint32_t)idx, store_volume_from_sink_cb, NULL);
+    if (o)
+        pa_operation_unref(o);
+    else {
```
stapelberg: nit: please be consistent about the parens. Either remove them from the else branch or add them to the if branch.

Also, if we can do error handling here, why not in every other place that starts a new pulseaudio operation?

src/pulse.c
```diff
@@ -0,0 +1,319 @@
...
+        pa_operation_unref(o);
+}
+
+static void subscribe_cb(pa_context *c, pa_subscription_event_type_t t,
+                         uint32_t idx, void *userdata) {
+    if ((t & PA_SUBSCRIPTION_EVENT_TYPE_MASK) != PA_SUBSCRIPTION_EVENT_CHANGE)
+        return;
+    pa_subscription_event_type_t facility =
+        t & PA_SUBSCRIPTION_EVENT_FACILITY_MASK;
+    if (facility == PA_SUBSCRIPTION_EVENT_SERVER) {
```
stapelberg: Since we now have multiple code paths, I think a `switch (facility)` would be easier to understand.

src/print_volume.c
```diff
@@ -51,13 +51,39 @@ void print_volume(yajl_gen json_gen, char *buffer, const char *fmt, const char *
...
+     * overriding the default sink */
+
+    if (!strncasecmp(device, "pulse", strlen("pulse"))) {
+        int sink_idx = device[5] == ':' ? atoi(device + 6) : -1;
+        int ivolume = pulse_initialize() ? volume_pulseaudio(sink_idx) : 0;
+        if (ivolume < 0)
+            ivolume = 0; /* negative result means error, stick to 0 */
+        outwalk = apply_volume_format(fmt, outwalk, ivolume);
+        goto out;
+    } else if (!strcmp(device, "default") && pulse_initialize()) {
```
stapelberg: Can you use `strcasecmp` instead of `strcmp` please?

stapelberg:
Thanks for all the work you’re putting into this. I feel like we got the functionality mostly nailed and are dealing only with code questions now.

As a heads-up, I’ll also want to see whether we can use http://clang.llvm.org/docs/ThreadSafetyAnalysis.html in order to make sure our mutex-related code is correct.

Watcom:
OK, I think all of the issues you pointed out were taken care of. The value to signify the default sink changed from -1 to UINT32_MAX and the type from int to uint32_t.

About TAILQ, I figured since it was ok to traverse the list of sinks to read a volume from the index, we might as well have one contiguous block which would be tighter than several tiny malloc'd blocks with the extra overhead of a doubly linked list. Of course we are talking about a few bytes here, the point being the realloc method doesn't seem more complicated to me than using TAILQ.

Very interesting and useful the Clang thread safety analysis feature. The page only shows C++ examples though, but I suppose there must be ways to make it work with C as well?

src/pulse.c
```diff
@@ -0,0 +1,303 @@
...
+            context_ready = 1;
+        } break;
+
+        case PA_CONTEXT_FAILED:
+            pulseaudio_error_log(c);
+            break;
+    }
+}
+
+static int find_sink_idx(uint32_t sink_idx) {
```
stapelberg: You’re not acquiring the pulse_mutex here, but in other places when accessing cached_volume.

include/i3status.h
```diff
@@ -14,10 +14,14 @@ enum { O_DZEN2,
 #include <yajl/yajl_version.h>
 #include <unistd.h>
 #include <string.h>
+#include <pthread.h>
+#include <stdint.h>
 
 #define BEGINS_WITH(haystack, needle) (strncmp(haystack, needle, strlen(needle)) == 0)
 #define max(a, b) ((a) > (b) ? (a) : (b))
 
+#define INVALID_SINK_INDEX UINT32_MAX
```
stapelberg: I think this should be DEFAULT_SINK_INDEX to be more clear about what its purpose is.

stapelberg:
> About TAILQ, I figured since it was ok to traverse the list of sinks to read a volume from the index, we might as well have one contiguous block which would be tighter than several tiny malloc'd blocks with the extra overhead of a doubly linked list. Of course we are talking about a few bytes here, the point being the realloc method doesn't seem more complicated to me than using TAILQ.

We’re talking about typically less than 10 mallocs. I don’t think it’s worth worrying about that, and I think using a TAILQ would make the code easier. Also note that modern libcs don’t translate every malloc() into an actual allocation, but have internal buffers. I really don’t think the memory layout is something that should concern us in such a case.

stapelberg:
> Very interesting and useful the Clang thread safety analysis feature. The page only shows C++ examples though, but I suppose there must be ways to make it work with C as well?

See https://llvm.org/bugs/show_bug.cgi?id=20403
With a bunch of type casts, you can actually compile i3status with -x c++: http://sprunge.us/aGUi

src/pulse.c
Watcom: The code is correct, though. I've added a comment stating the mutex should be locked prior to calling find_sink_idx().

Watcom:
TAILQ it is, then.

include/i3status.h
Watcom: Done.

src/pulse.c
```diff
@@ -0,0 +1,304 @@
...
+
+/*
+ *  pulse_initialize
+ *
+ *  detect and, if necessary, initialize the PulseAudio API
+ */
+int pulse_initialize(void) {
+    if (!initialized) {
+        initialized = 1;
+        TAILQ_INIT(&cached_volume);
```
stapelberg: Why do you initialize the tailq again? This line is equivalent to what the TAILQ_HEAD_INITIALIZER does.

src/pulse.c
```diff
@@ -0,0 +1,304 @@
...
+        } break;
+
+        case PA_CONTEXT_FAILED:
+            pulseaudio_error_log(c);
+            break;
+    }
+}
+
+/* must be called with pulse_mutex locked */
+static const indexed_volume_t *find_sink_idx(uint32_t sink_idx) {
```
stapelberg: Now that we’ve switched to tailq, we can refactor this to make it simpler:

Rename it to `find_volume_by_idx` and make it always return a const indexed_volume_t*, i.e. make it allocate one if it cannot be found. Then, you can get rid of `store_wanted_sink` and `volume_from_sink_idx` entirely by just calling `find_volume_by_idx` instead.

src/pulse.c
```diff
@@ -0,0 +1,304 @@
...
+typedef struct indexed_volume_s {
+    uint32_t idx;
+    int volume;
+    TAILQ_ENTRY(indexed_volume_s) entries;
+} indexed_volume_t;
+
+static pa_threaded_mainloop *main_loop = NULL;
+static pa_context *context = NULL;
+static pa_mainloop_api *api = NULL;
+static int context_ready = 0;
```
stapelberg: Here and below, can you please use bools instead of ints when you’re only putting 0 and 1 into them? Then you can use true/false, which is easier to read.

src/pulse.c
```diff
@@ -0,0 +1,304 @@
...
+        pa_threaded_mainloop_lock(main_loop);
+        get_sink_info(context, sink_idx);
+        pa_threaded_mainloop_unlock(main_loop);
+        return 0;
+    }
+    return vol;
+}
+
+/*
+ *  pulse_initialize
```
stapelberg: Here and elsewhere, can you please remove the function name from the comment for consistency with the rest of our code base?

src/print_volume.c
```diff
@@ -51,13 +51,40 @@ void print_volume(yajl_gen json_gen, char *buffer, const char *fmt, const char *
...
+    } else if (!strcasecmp(device, "default") && pulse_initialize()) {
+        /* no device specified or "default" set */
+        int ivolume = volume_pulseaudio(DEFAULT_SINK_INDEX);
+        if (ivolume >= 0) {
+            outwalk = apply_volume_format(fmt, outwalk, ivolume);
+            goto out;
+        }
+        /* negative result means error, fail PulseAudio attempt */
+    }
+/* If some other device was specified or PulseAudio is not detected,
```
stapelberg: nit: this comment seems to not be correctly formatted.

i3status.c
```diff
@@ -688,7 +692,17 @@ int main(int argc, char *argv[]) {
          * that we start with :00 on every new minute. */
         struct timeval current_timeval;
         gettimeofday(&current_timeval, NULL);
-        struct timespec ts = {interval - 1 - (current_timeval.tv_sec % interval), (10e5 - current_timeval.tv_usec) * 1000};
-        nanosleep(&ts, NULL);
+        struct timespec ts;
+
+        clock_gettime(CLOCK_REALTIME, &ts);
```
stapelberg: Can you explain a little bit what you’re changing here? It seems a bit odd to me that both the `gettimeofday` and the `clock_gettime` calls are necessary.

stapelberg:
Thanks for your continued patience on this. I think in one or two rounds we’ll end up with high-quality code that’s consistent with the rest of our code base. I’ll be happy to merge it then :).

src/pulse.c
Watcom: Sorry, my mistake. Fixed.

src/pulse.c
Watcom: Actually, only the main thread can store new sink indices to the list. The PulseAudio thread only sees wether a sink's volume should be saved or not based on it existing there. This way only the volumes of the sinks specified by the user need to be saved. As a consequence it's not possible to unify them in the way you describe.

src/print_volume.c
Watcom: Hmm, I fixed this twice before. It seems clang-format doesn't want it aligned to the code above, but to the code below. But it only aligns the first line, so I aligned the second line to the left this time.

i3status.c
Watcom: You are right in saying it is odd, it's more complicated than it could be. I just made it a lot simpler and got rid of `gettimeofday()`.

Watcom:
>  Thanks for your continued patience on this. I think in one or two rounds we’ll end up with high-quality code that’s consistent with the rest of our code base. I’ll be happy to merge it then :).

No problem at all. I also benefit from this as a user :).

i3status.c
```diff
@@ -686,9 +690,14 @@ int main(int argc, char *argv[]) {
...
          * seconds. We also align to 60 seconds modulo interval such
          * that we start with :00 on every new minute. */
-        struct timeval current_timeval;
-        gettimeofday(&current_timeval, NULL);
-        struct timespec ts = {interval - 1 - (current_timeval.tv_sec % interval), (10e5 - current_timeval.tv_usec) * 1000};
-        nanosleep(&ts, NULL);
+        struct timespec ts;
+        clock_gettime(CLOCK_REALTIME, &ts);
+        ts.tv_sec += interval - (ts.tv_sec % interval);
+        ts.tv_nsec = 0;
```
stapelberg: Now it seems like you’re missing the tv_nsec part, i.e. you’ll not wake exactly on the full second, right? (see the comment above)

src/print_volume.c
stapelberg: I think adding an empty line before and after the comment will make clang-format align it with the code above.

src/pulse.c
stapelberg: Hm, but we could change it so that all sink volumes are saved and then do the unification, right? I think the additional memory usage is negligible in this case.

Watcom: OK, I see your point. I went a step further and inlined the whole thing since it made it even simpler.

src/print_volume.c
Watcom: No, that didn't work. The #ifdef below seems to confuse clang-format.

i3status.c
Watcom: Actually no, it behaves just like before. That includes the ":00 on every new minute" part. You can check that using a 5-second interval. The reason it looks simpler is that while `nanosleep` expects a relative time (i.e. a delay), `pthread_cond_timedwait` expects an absolute time, so it's actually easier to achieve the same result.

src/pulse.c
```diff
@@ -0,0 +1,264 @@
...
+            pa_strerror(pa_context_errno(c)));
+}
+
+static bool pulseaudio_free_operation(pa_context *c, pa_operation *o) {
+    if (o)
+        pa_operation_unref(o);
+    else
+        pulseaudio_error_log(c);
+    /* return false if the operation failed */
+    return !!o;
```
stapelberg: nit: `return o;` is enough here and consistent with your if statement on line 35.

src/pulse.c
```diff
@@ -0,0 +1,264 @@
...
+    pthread_mutex_lock(&pulse_mutex);
+    const indexed_volume_t *entry;
+    TAILQ_FOREACH(entry, &cached_volume, entries) {
+        if (entry->idx == sink_idx) {
+            int vol = entry->volume;
+            pthread_mutex_unlock(&pulse_mutex);
+            return vol == BLANK_VOLUME ? 0 : vol;
+        }
+    }
+    /* index not found, store it so we don't end up requesting the volume
```
stapelberg: Given that the interval is typically ≥ 1 second, it is _extremely_ unlikely that the pulseaudio callback will not be called before the next call of `volume_pulseaudio`. And even if it’s not called, all that happens is another (unnecessary) call to `get_sink_info`, right?

I suggest we replace the remainder of this function with just a call to `get_sink_info`, and then you can also inline `allocate_volume_entry` into `save_volume`.

src/pulse.c
```diff
@@ -0,0 +1,264 @@
...
+            return true;
+        }
+    }
+    /* index not found, store it */
+    allocate_volume_entry(sink_idx, new_volume);
+    pthread_mutex_unlock(&pulse_mutex);
+    return true;
+}
+
+static void wake_main_thread(void) {
```
stapelberg: Since this function is called only once, inline it please.

src/pulse.c
```diff
@@ -0,0 +1,264 @@
...
+/*
+ * conditionally save the volume for the specified sink index
+ * returning true if the value was changed
+ */
+static bool save_volume(uint32_t sink_idx, int new_volume) {
+    pthread_mutex_lock(&pulse_mutex);
+    indexed_volume_t *entry;
+    TAILQ_FOREACH(entry, &cached_volume, entries) {
+        if (entry->idx == sink_idx) {
+            if (new_volume == entry->volume) {
```
stapelberg: I suggest:

```c
const bool changed = (new_volume != entry->volume);
entry->volume = new_volume;
pthread_mutex_unlock(&pulse_mutex);
return changed;
```

This reduces two very similar code paths to one and eliminates a conditional (at the expense of a write, but since this is only called when the volume changes, that’s acceptable).

i3status.c
stapelberg: Ah, I see. Can you update the comment below, then? You’re currently saying `pthread_cond_timedwait` is equivalent to a `nanosleep` call, but based on this observation, it clearly isn’t :).

Watcom:
Done, pushed commit.

src/pulse.c
```diff
@@ -0,0 +1,245 @@
...
+
+static void store_default_sink_cb(pa_context *c,
+                                  const pa_sink_info *i,
+                                  int eol,
+                                  void *userdata) {
+    if (i) {
+        if (default_sink_idx != i->index) {
+            /* default sink changed? */
+            default_sink_idx = i->index;
+            get_sink_info(c, default_sink_idx);
```
stapelberg: Instead of calling `get_sink_info`, you could also just call `store_volume_from_sink_cb` to save another round-trip, right?

Watcom: Indeed!

stapelberg:
Very cool. Thanks again!

Watcom:
Thank you!

PR #8, opened 2015-03-13
Add contrib/net-speed, a reimplementation of contrib/measure-net-speed.bash

azuwis:
A reimplementation of contrib/measure-net-speed.bash.

- Single script
- Compatible with most POSIX shells, tested with bash, dash, busybox ash
- Auto detect interfaces
- Does not write temp files

Screenshot:

screenshot


breunigs:
Very nice! Would it be possible to add displaying "6M" instead of "5931K"?

azuwis:
@breunigs Done.

screenshot

PR #9, opened 2015-03-17
Added setting for 'format_load' to set different format when load is over threshold

jefvel:
For displaying a warning icon when over the threshold or something.

### DIFF
--- a/i3status.c
+++ b/i3status.c
@@ -386,6 +386,7 @@ int main(int argc, char *argv[]) {
 
         cfg_opt_t disk_opts[] = {
                 CFG_STR("format", "%free", CFGF_NONE),
+                CFG_STR("format_not_mounted", NULL, CFGF_NONE),
                 CFG_STR("prefix_type", "binary", CFGF_NONE),
                 CFG_STR("threshold_type", "percentage_avail", CFGF_NONE),
                 CFG_FLOAT("low_threshold", 0, CFGF_NONE),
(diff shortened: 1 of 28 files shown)


stapelberg:
Please also update the manpage.

Please rebase your commits, so that the “Merge branch 'master'” one disappears.

jefvel:
I updated the manpage.
I'm not sure how to rebase this correctly, I've never done that before.

i3status.c
```diff
@@ -362,6 +362,7 @@ int main(int argc, char *argv[]) {
 
     cfg_opt_t load_opts[] = {
         CFG_STR("format", "%1min %5min %15min", CFGF_NONE),
+        CFG_STR("format_load", "%1min %5min %15min", CFGF_NONE),
```
stapelberg: I think it would be better to use `NULL` instead of replicating the normal format string. Then you can check for `NULL` and fall back on `format` in `print_load()`. This is important because it preserves the current behavior for people who changed `format`.

stapelberg:
See https://help.github.com/articles/about-git-rebase/ for how to rebase.

i3status.c
```diff
@@ -621,7 +622,13 @@ int main(int argc, char *argv[]) {
 
             CASE_SEC("load") {
                 SEC_OPEN_MAP("load");
-                print_load(json_gen, buffer, cfg_getstr(sec, "format"), cfg_getfloat(sec, "max_threshold"));
+                const char* format_load = cfg_getstr(sec, "format_load");
+                const char* format = cfg_getstr(sec, "format");
+                if(format_load == NULL) {
```
chrko: I don't understand why you're checking at this point wether format_load is NULL or not. If not you are calling the parameter format_load with format. But this could be easily done in the function itself.

src/print_load.c
```diff
@@ -18,11 +18,18 @@ void print_load(yajl_gen json_gen, char *buffer, const char *format, const float
     if (getloadavg(loadavg, 3) == -1)
         goto error;
 
-    for (walk = format; *walk != '\0'; walk++) {
+    if (loadavg[0] >= max_threshold) {
```
chrko: So this should be the place to my opinion. So my proposal is to change it to the following:
```c
if (format_down != NULL && loadavg[0] >= max_threshold) {
    walk = format_load;
} else {
    walk = format;
}
```

stapelberg:
@jefvel Sorry for the lack of communication on my part. You’ve not replied to my comments, and GitHub doesn’t send an email notification for when you just push commits. Please add a comment the next time you change something :).

Airblader:
@jefvel Are you still working on this?

Airblader:
I'm going to close this PR now since there hasn't been activity in far over half a year.

PR #12, opened 2015-03-23
Pass the "instance" JSON property key for tztime blocks

### LINKED ISSUE #10: pass instance key on tztime click events
Hi,

Currently, when i3status passes a click event on a **tztime module** to an external program, it does not include its instance name (which is mendatory for tztime).

    {'y': 9, 'x': 2256, 'button': 1, 'name': 'tztime'}

Could you please make it so that i3status passes an `instance` key with the name of the tztime module being clicked ?

Please also make sure it does so on `time` modules with instance names.

Airblader:
fixes #10

### DIFF
--- a/i3status.c
+++ b/i3status.c
@@ -627,13 +627,13 @@ int main(int argc, char *argv[]) {
 
             CASE_SEC("time") {
                 SEC_OPEN_MAP("time");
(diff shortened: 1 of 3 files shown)


Airblader:
@stapelberg Out of curiosity, when will i3status be released? Looks like it's been a while.

PR #14, opened 2015-03-27
conf: set output_format to i3bar

acrisci:


stapelberg:
Why? It gets auto-detected.

acrisci:
I did not know that. I was testing it on the command line and thought it was giving plaintext output to i3bar. Auto-detect is good.

PR #15, opened 2015-03-28
refactoring to avoid code duplication

Watcom:
Refactoring to avoid code duplication.

### DIFF
--- a/src/print_volume.c
+++ b/src/print_volume.c
@@ -27,6 +27,26 @@
 #include "i3status.h"
 #include "queue.h"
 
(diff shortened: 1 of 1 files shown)


src/print_volume.c
```diff
@@ -27,6 +27,34 @@
 #include "i3status.h"
 #include "queue.h"
 
+/*
+ * apply_volume_format
```
stapelberg: Here and below, can you remove the comment above the function please? It doesn’t add any value.

PR #17, opened 2015-04-02
Adding format_down option to path_exists and run_watch

chrko:
According to #1 the option format_down (to hide the output) is missing on path_exists. This also add the option to run_watch.
Do not change behavior without editing config, so no user will claim the new optional behaviour!

### DIFF
--- a/man/i3status.man
+++ b/man/i3status.man
@@ -239,7 +239,7 @@ implies no coloring at all.
 
 You can define a different format with the option "format_not_mounted"
 which is used if the path is not a mount point. So you can just empty
-the output for the given path with adding »format_not_mounted=""«
+the output for the given path with adding +format_not_mounted=""+
 to the config section.
 
 *Example order*: +disk /mnt/usbstick+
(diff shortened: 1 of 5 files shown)


man/i3status.man
```diff
@@ -259,6 +259,8 @@ to the config section.
 Expands the given path to a pidfile and checks if the process ID found inside
 is valid (that is, if the process is running). You can use this to check if
 a specific application, such as a VPN client or your DHCP client is running.
+It also exists an option "format_down". You can hide the output with
```
stapelberg: “It also exists” is a german idiom that doesn’t translate directly to english. Use “There also is” instead. Also, please replace the quotes with +, i.e. `+format_down=""+` to be consistent with the rest of this manpage.

stapelberg:
Please squash your commits into one commit so that we have a clean history.

Also, please add “fixes #1” to the commit description so that the issue will automatically be closed.

chrko:
Squashed it.

PR #19, opened 2015-04-03
add readline from file status option

jbremer:
Allows one to do the following in the ``~/.i3status.conf``:

```
order += "readline HOSTS"

...

readline HOSTS {
    path = "/etc/hosts"
}
```

It handles invalid non-printable characters and doesn't emit multiple consecutive whitespaces. For me this is useful so I can write a cronjob which updates a textfile every now and then, and then I can see this updated value in the i3status bar.

Jurriaan

### DIFF
--- a/i3status.c
+++ b/i3status.c
@@ -298,6 +298,14 @@ int main(int argc, char *argv[]) {
         CFG_STR("format", "%title: %status", CFGF_NONE),
         CFG_STR("format_down", NULL, CFGF_NONE),
         CFG_CUSTOM_ALIGN_OPT,
+        CFG_CUSTOM_MIN_WIDTH_OPT,
+        CFG_END()};
+
+    cfg_opt_t readline_opts[] = {
+        CFG_STR("path", NULL, CFGF_NONE),
+        CFG_STR("format", "%title: %status", CFGF_NONE),
+        CFG_STR("format_down", NULL, CFGF_NONE),
+        CFG_CUSTOM_ALIGN_OPT,
         CFG_CUSTOM_COLOR_OPTS,
         CFG_CUSTOM_MIN_WIDTH_OPT,
         CFG_END()};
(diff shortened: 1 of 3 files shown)


stapelberg:
> It handles invalid non-printable characters and doesn't emit multiple consecutive whitespaces

This sort of logic is why we won’t accept such a feature. We’ve had multiple requests in the past, and it was always necessary to do some sort of light postprocessing of data.

Sorry for the wasted effort you put into this pull request (you’re welcome to ask whether features will be accepted by creating an issue next time).

I recommend you instead wrap i3status as described in http://i3wm.org/i3status/manpage.html#_external_scripts_programs_with_i3status — that way, you’ll have much more flexible postprocessing.

jbremer:
This is pretty much the only feature I'm missing in i3, so I'm not sure if I'll have any other suggestions later. The idea behind ``i3status`` is that it has a low CPU overhead - mixing it with an ugly bash script kind of defeats the point ;)

The reason I decided that a file would be useful is because you don't want an asynchronous command to occur when refreshing the i3status status - reading a file is quick. In fact, half of the features that are supported by i3status require *some sort of light postprocessing of data*. Even more, they're OS-dependant (e.g., battery, cpu temperature, cpu usage, etc) and thus require the same logic up to three times or so.

Long story short, I'd barely consider this as processing data at all - it's merely *representing* a string which may contain arbitrary 8-bit characters. As you're probably not going to merge it I'll just keep it for myself :)

Other than that, i3 is great, thanks!

stapelberg:
> The idea behind i3status is that it has a low CPU overhead - mixing it with an ugly bash script kind of defeats the point ;)

If you constrain yourself to shell builtins, there will not be a measurable CPU overhead with any modern shell.

Alternatively, you can use any higher level scripting language to achieve the same thing. The contrib/ directory has examples in Perl and Python.

> In fact, half of the features that are supported by i3status require some sort of light postprocessing of data. Even more, they're OS-dependant (e.g., battery, cpu temperature, cpu usage, etc) and thus require the same logic up to three times or so.

Sorry, I should have been clearer. The point is that the sort of postprocessing you want to do is not _exactly_ the sort of postprocessing some other user wants to do, and so on. Everyone has their own little use-case that almost works, if only i3status supported *[insert postprocessing feature here]* on plain text files. That inevitably leads to growth and complexity, which we want to avoid.

acrisci:
For this level of customization, I would suggest trying my new statusline generator, [i3-dstatus](https://github.com/acrisci/i3-dstatus), which supports updating the statusline from a cron in bash.

PR #21, opened 2015-04-03
Added support for separator_block_width

### LINKED ISSUE #20: [enhancement] sep_block_width
i3status should support changing the `sep_block_width` property in its config:

````
general {
    sep_block_width = 9
}
```

Airblader:
The general section of the config can now accept a 'separator_block_width' which will be used if the output_format is i3bar.
The value will only be included in the JSON output if it differs from 9, which is the default in i3.

fixes #20

### DIFF
--- a/include/i3status.h
+++ b/include/i3status.h
@@ -88,32 +88,37 @@ enum { O_DZEN2,
         }                                                                              \
     } while (0)
 
-#define SEC_CLOSE_MAP                                                                                    \
-    do {                                                                                                 \
-        if (output_format == O_I3BAR) {                                                                  \
-            char *_align = cfg_getstr(sec, "align");                                                     \
-            if (_align) {                                                                                \
-                yajl_gen_string(json_gen, (const unsigned char *) "align", strlen("align"));             \
-                yajl_gen_string(json_gen, (const unsigned char *)_align, strlen(_align));                \
-            }                                                                                            \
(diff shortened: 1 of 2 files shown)


include/i3status.h
```diff
@@ -88,32 +88,37 @@ enum { O_DZEN2,
...
+                }                                                                                                            \
+            }                                                                                                                \
+            const char *_sep = cfg_getstr(cfg_general, "separator");                                                         \
+            if (strlen(_sep) == 0) {                                                                                         \
+                yajl_gen_string(json_gen, (const unsigned char *) "separator", strlen("separator"));                         \
+                yajl_gen_bool(json_gen, false);                                                                              \
+            }                                                                                                                \
+            const int _sep_block_width = cfg_getint(cfg_general, "separator_block_width");                                   \
+            if (_sep_block_width != 9) {                                                                                     \
```
Airblader: I don't really like the dependency on i3bar's default here since this are two different tools. However, I also don't think it's good to dump this all the time. It'd be good if we could avoid a situation like this altogether by having a separate config file for i3bar rather than forcing all tools to implement this themselves.

Airblader:
Thinking a bit more about this, I'm not so sure this is a good idea. It's a per-block property and i3status now handles it as a global property. In this specific case, https://github.com/i3/i3/pull/1634 would take care of this anyway (though not in the same way).

PR #23, opened 2015-04-04
Basic XDG support and minor tweaks to the net-speed script

Deiz:


stapelberg:
What are the “minor tweaks” you’re describing? I can only see the XDG support in this pull request.

Deiz:
"Minor tweaks" refers to lines 59 to 63, just ensuring that the conf exists before trying to grep it.

PR #26, opened 2015-04-16
Changed battery info output to HH:MM for consistency accross all platforms

nhelke:
Minutes left is not a useful measure of battery life (except on laptops with tiny batteries which last < 60 minutes). HH:MM is more friendly and inline with what other platforms show.

### DIFF
--- a/src/print_battery_info.c
+++ b/src/print_battery_info.c
@@ -245,7 +245,7 @@ void print_battery_info(yajl_gen json_gen, char *buffer, int number, const char 
         minutes = remaining;
         hours = minutes / 60;
         minutes -= (hours * 60);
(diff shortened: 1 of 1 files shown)


stapelberg:
Please use clang-format to format your code (see the failed travis build).

Airblader:
@stapelberg Ping
@nhelke You need to write a comment after pushing changes to generate a notification to Michael.

stapelberg:
@nhelke Can you please squash your commits (using interactive rebase, followed by a forced push) so that our history remains clean?

nhelke:
Of course I can. I'm in the outback at the moment, but I will do it next
week when I get back to a computer. (Lesson learned @Airblader, this time I
will ping @stapelberg after I have pushed.)

nhelke:
@stapelberg I have fixed up the two commits.

PR #28, opened 2015-04-19
PulseAudio muted volume support

Watcom:
PulseAudio muted volume support, which was missing in the original implementation.

PR #29, opened 2015-04-19
if the volume device is "alsa", force using ALSA PulseAudio emulation

Watcom:
The ability to force using the "pulse" ALSA device (using PulseAudio through ALSA emulation) was missing. This patch makes that possible by specifying the device "alsa".

### DIFF
--- a/man/i3status.man
+++ b/man/i3status.man
@@ -447,6 +447,8 @@ default is used. If the device string is missing or is set to "default",
 PulseAudio will be tried if detected and will fallback to ALSA (Linux)
 or OSS (FreeBSD/OpenBSD).
 
+To force PulseAudio using ALSA emulation, use "alsa" as the device name.
+
 *Example order*: +volume master+
 
 *Example format*: +♪: %volume+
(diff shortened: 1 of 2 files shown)


man/i3status.man
```diff
@@ -447,6 +447,8 @@ default is used. If the device string is missing or is set to "default",
 PulseAudio will be tried if detected and will fallback to ALSA (Linux)
 or OSS (FreeBSD/OpenBSD).
 
+To force PulseAudio using ALSA emulation, use "alsa" as the device name.
```
stapelberg: I don’t understand this sentence. Which API will be used when specifying “alsa” and why is it necessary to have this special case?

Watcom: You could already force using the ALSA API by using any device name other than "pulse", "default" or missing value. The problem is that the ALSA device name for the PulseAudio backend is also "pulse", so there's no way to force using it. This is one way to address this problem.

stapelberg: Okay, but why would you want to enforce using ALSA when you’re using PulseAudio via ALSA’s PulseAudio backend anyway? I don’t see the point in that.

Watcom: I suppose to mimic the old behaviour, without the thread wakes, etc. A vague reason, admittedly.

monstermunchkin: The reason I suggested it was to mimic the old behaviour since I never had problems with muted volume then. #28 however fixed that problem.

Watcom: OK, closing pull request.

PR #34, opened 2015-05-28
Print current keyboard layout

idrarig:
I frequently switch keyboard layouts so I find it useful to know which one is current from the status bar.  With this patch I just add `order += "keyboard_layout"` to get, for instance, `L: us_intl` added to the status bar.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,7 @@ CFLAGS+=-Iinclude
 LIBS+=-lconfuse
 LIBS+=-lyajl
 LIBS+=-lpulse
+LIBS+=-lX11 -lxkbfile
 
 VERSION:=$(shell git describe --tags --abbrev=0)
 GIT_VERSION:="$(shell git describe --tags --always) ($(shell git log --pretty=format:%cd --date=short -n1))"
--- a/i3status.c
+++ b/i3status.c
@@ -410,6 +410,14 @@ int main(int argc, char *argv[]) {
         CFG_CUSTOM_MIN_WIDTH_OPT,
         CFG_END()};
 
+    cfg_opt_t keyboard_layout_opts[] = {
+        CFG_STR("format", "L: %s", CFGF_NONE),
+        CFG_CUSTOM_ALIGN_OPT,
+        CFG_CUSTOM_COLOR_OPTS,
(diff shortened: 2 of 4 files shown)


Airblader:
Why aren't you using xcb-xkb? The i3 suite is generally written in XCB and not Xlib, unless there's a good reason to I think we should keep it that way.

idrarig:
@Airblader I think when I did this (it was months ago, I just rebased on master and pushed) I ltrace-d some program to see how I could get that information and then I looked at the sources of said program to see how I could do the same in i3status.  I'd be happy to look into another implementation.  Could you give me a pointer where to start with xcb-xkb?

Airblader:
Not really, sorry, I never worked with xkb before. I'm sure @stapelberg can say somehting about it, though.

idrarig:
Don't worry, after all, the source is out there :-). Thanks!

stapelberg:
In light of pull request #53 (which adds something similar), I thought a bit about this, and I don’t think displaying the current keyboard layout is a useful addition to i3status. You’ll figure out immediately in which keyboard layout you are when you press a key and it doesn’t have the expected result.

puiterwijk:
Note that I don't entirely agree with your last remark, as it might be tricky to depend on the first keystroke to let users determine their keymap:

- If you are entering a password, you may enter it incorrectly multiple times before trying to switch keyboard layout
- If you are using a program mainly controlled by keyboard, you might have initiated an action before noticing you had the wrong layout selected (like having selected "n" to a question while you wanted another response)
- If you are using an IME, the first keystroke(s) might not show up or do anything until you hit enter to confirm the input, depending on where your text cursor is currently located, which might be very confusing.

True, all of these are still possible with a textual display, but it is possible to train oneself to look at the status before entering a password or executing a command.

Airblader:
I agree that it makes sense to see the keyboard layout and I'm pretty sure I've seen this on most other desktops as well in some form or another. However, that doesn't yet mean it should be in i3status. I think there are keyboard layout indicator tray widgets out there which could be used here.

You'll also always have the option to move away from i3status, which deliberately doesn't try to be the most extensive and versatile tool for i3bar.

PR #36, opened 2015-05-30
Add missing dependency information

ambihelical:
The following additional packages were needed to build:
libxml2-utils, xsltproc

### DIFF
--- a/README
+++ b/README
@@ -21,9 +21,11 @@ i3status has the following dependencies:
  • libcap2-bin (for getting network status without root permissions)
  • asciidoc (only for the documentation)
  • libpulse-dev (for getting the current volume using PulseAudio)
+ • libxml2-utils (for xmllint)
+ • xsltproc
 
 On debian-based systems, the following line will install all requirements:
-apt-get install libconfuse-dev libyajl-dev libasound2-dev libiw-dev asciidoc libcap2-bin libpulse-dev
+apt-get install libconfuse-dev libyajl-dev libasound2-dev libiw-dev asciidoc libcap2-bin libpulse-dev libxml2-utils xsltproc
 
  ┌────────────────────────────┐
  │ Upstream                   │


stapelberg:
On which distribution is that? We’re not using these libraries directly in the i3status build process, so asciidoc should pull them in.

ambihelical:
Mint 17.1

ambihelical:
Looking into it a little further, libxml2-utils is a recommended package, not a dependency.  xsltproc is a dependency of at least one recommended package as well.   My system setup scripts don't install recommended packages, so I guess this is technically a problem with my setup.

Airblader:
If a build fails due to a missing package, that package should be a hard dependency, not a recommendation. Michael's point is just that these two dependencies aren't dependencies of i3 itself, but of asciidoc which in turn is a dependency of i3.

ambihelical:
Just to be clear, I was refering to asciidoc's recommended packages in my previous comment.   Asciidoc is not one of the packages I install with `apt-get install --no-install-recommends`, but it may have been pulled in as a dependency by some other package using that option. I can't really say at this point, but it would explain why those packages were missing.

stapelberg:
Yeah, that’s what I suspected.

https://www.debian.org/doc/debian-policy/ch-relationships.html#s-binarydeps defines:

> Recommends
> This declares a strong, but not absolute, dependency.
> The Recommends field should list packages that would be found together with this one in all but unusual installations.

The way I interpret this is that users should not disable Recommends unless they are willing to accept situations like the one you found yourself in :).

Unfortunately, adding these as dependencies doesn’t solve the issue at hand in a reliable way: it could be that asciidoc gains another Recommends in future versions, and then we’d need to play catch-up. Also, asciidoc might drop the Recommends at some point, and then we are pulling in an unnecessary dependency.

In conclusion, I don’t think adding these dependencies is a good idea.

PR #37, opened 2015-06-02
SIGUSR1 fix regains capability to force status update. Maybe just tem…

holoubekm:
Attempt to fix regression in the reaction to signals. SIGUSR1 can be used to force the update again. Re-enables prompt reaction to SIGTERM signal. I think this capability should be preservet, not everything can be rewritten into async model like print_volume.

Re-enables this kind of commands:
`bindsym XF86AudioRaiseVolume exec --no-startup-id ~/.i3/raise_vol.sh && killall -SIGUSR1 i3status`

### DIFF
--- a/i3status.c
+++ b/i3status.c
@@ -74,14 +74,21 @@ pthread_mutex_t i3status_sleep_mutex = PTHREAD_MUTEX_INITIALIZER;
  */
 void fatalsig(int signum) {
     exit_upon_signal = true;
+    // Broadcast ensures swift reaction times
+    // pthread_mutex_lock(&i3status_sleep_mutex);
+    pthread_cond_broadcast(&i3status_sleep_cond);
+    // pthread_mutex_unlock(&i3status_sleep_mutex);
 }
 
 /*
- * Do nothing upon SIGUSR1. Running this signal handler will nevertheless
- * interrupt nanosleep() so that i3status immediately generates new output.
+ * Here, we wake up the main thread sleeping. 
+ * This restores the lost capability to react to the SIGUSR1 signal.
  *
  */
 void sigusr1(int signum) {
+    // pthread_mutex_lock(&i3status_sleep_mutex);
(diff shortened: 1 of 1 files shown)


stapelberg:
Looks good, but can you squash the commits please so that our commit history stays clean?

holoubekm:
Mission accomplished.

holoubekm:
You're right man! I didn't realize this before. As far as i understand threading we have two possibilities. Remove the mutex lock, with a low chance of missed wakeup or implement `signalfd` and use polling thread waiting for signal. What do you suppose?

Watcom:
Removing the lock/unlock will avoid a deadlock, but to be pedantic and interpreting the man page to the letter, none of the pthread functions are async-signal-handler-safe. BTW, `signalfd` is Linux-specific, so it would need some conditional compilation.

I haven't really looked further into this problem so I honestly don't have an opinion about it. Maybe just removing the lock/unlock may be OK in practice, I don't know.

holoubekm:
Yes, it would remove the deadlock problem, on the other hand results would be slightly unpredictable depending on the scheduler. This seems to me be the easy way, otherwise it would be `kqueue` vs `epoll` conditional compilation.

I have just removed lock/unlock as the cheap solution.

stapelberg:
I think a message by @Watcom is missing in this thread…? I only see @holoubekm’s “You're right man!” without context. @Watcom, what did you point out specifically?

Watcom:
@stapelberg Sorry, I accidently commented on the commit tab and the message went away when he changed the code. I pointed out the code wasn't safe according to the pthread man page:

> Async-signal Safety

> The mutex functions are not async-signal safe. What this means is that they should not be called from a signal handler. In particular, calling pthread_mutex_lock or pthread_mutex_unlock from a signal handler may deadlock the calling thread.

Airblader:
I'm going to close this since there hasn't been activity. Also seems to have been fixed by other PRs in the meantime, if I'm not mistaken (?). Either way, closing.

PR #38, opened 2015-06-02
Added brightness support.

holoubekm:
Module adds brightness support for i3status. Expecting two files, one with maximum value and one with current value of the brightness. Module has support for numeric and symbolic output. May be compile with `-lphread` for async mode. Thread then observes the brightness input file and occasionally forces the main one to be updated.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,9 @@ ifndef SYSCONFDIR
   endif
 endif
 
+# Support for conditional async build
+ASYNC=1
+
 CFLAGS+=-Wall -Wshadow -Wpointer-arith -Wcast-qual -Wsign-compare
 CFLAGS+=-g
 CFLAGS+=-std=gnu99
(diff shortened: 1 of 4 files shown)


stapelberg:
I don’t understand the use-case for this feature yet, can you elaborate?

In my point of view, brightness is something you perceive and don’t need to be told about. I.e., if your screen is not bright enough, you increase brightness, and if you think it’s too bright, you decrease brightness. Why would you ever need to look at i3status to know how bright your screen is?

holoubekm:
This module helps me to quickly estimate the rough power consumption, Or
maybe it's just my obsession for numbers and knowledge of system status.

stapelberg:
Thanks for clarifying. Given that i3status already displays the power consumption, I’ll reject this pull request. For breaking down power consumption, if necessary, there are better tools, e.g. `powertop`.

PR #40, opened 2015-07-06
Mac OS X compatibility changes

afh:
Some minor changes to make i3status compile under OS X.

:warning: I'm uncertain whether the additional call to `gettimeofday` in https://github.com/i3/i3status/commit/2113737107c537e9cbc1a40d2d8d5fac5152e03e is really necessary or whether the call to `gettimeofday`in line 563 is enough.
What's your opinion?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,8 +47,8 @@ ifeq ($(OS),OpenBSD)
 LIBS+=-lossaudio
 endif
 
-ifeq ($(OS), NetBSD)
-LIBS+= -lprop
+ifeq ($(OS),NetBSD)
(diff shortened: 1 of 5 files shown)


stapelberg:
> I'm uncertain whether the additional call to gettimeofday in 2113737 is really necessary or whether the call to gettimeofdayin line 563 is enough.
> What's your opinion?

I think the second call is necessary, since the first one is run _before_ all the stats are gathered, which might cause enough skew to be noticeable.

afh:
Great! Thanks for merging.

PR #42, opened 2015-07-15
Use color_degraded if battery charge is above 100%.

KarboniteKream:
The battery on my laptop has some issues and battery charge sometimes goes above 100% (upower shows energy-full higher than energy-full-design).

This commit colors the output yellow (color_degraded), which notifies the user that the battery might be having some issues.

### DIFF
--- a/src/print_battery_info.c
+++ b/src/print_battery_info.c
@@ -172,6 +172,9 @@ void print_battery_info(yajl_gen json_gen, char *buffer, int number, const char 
             if (strcasecmp(threshold_type, "percentage") == 0 && percentage_remaining < low_threshold) {
                 START_COLOR("color_bad");
                 colorful_output = true;
+            } else if (strcasecmp(threshold_type, "percentage") == 0 && percentage_remaining > 100.0) {
+                START_COLOR("color_degraded");
+                colorful_output = true;
             } else if (strcasecmp(threshold_type, "time") == 0 && seconds_remaining < 60 * low_threshold) {
                 START_COLOR("color_bad");
                 colorful_output = true;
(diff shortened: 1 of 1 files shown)


Airblader:
I'm not so sure about this change because it sounds more like a problem of the system, but if Michael does want to merge it: as t the very least the documentation is missing.

stapelberg:
I don’t quite understand why the user would need to care about the battery delivering _more_ capacity than it’s supposed to? Can you elaborate on what a user might do in response?

KarboniteKream:
When I got this battery, everything was working fine, but after a while, I noticed that both i3status and upower started to report this "issue", so I would call this a malfunction, since this never occurred with the previous battery.

I implemented this so it was easier for me to notice when the capacity was over 100%, as I was trying to figure out what might be wrong, and I thought this might help others, at least to let them know that their battery might be malfunctioning. If you think this isn't useful, you can close the pull request.

stapelberg:
I still don’t understand what you mean when you say “malfunction”. If your battery capacity is reported as more than 100%, that means either it delivers more capacity than the vendor deems usual (which is the normal state for new batteries in their first few cycles), the vendor might have specified a wrong number, the electronics are mis-calibrated for some reason, etc.

Neither of these cases seems like it requires any user action, so I wouldn’t know why the user should care?

KarboniteKream:
The coloring is only meant to be informative. I only created this PR, because I thought it might be useful for someone else.

stapelberg:
I’m going to close this now, because I still don’t see an improvement in doing this. The degraded/bad colors are meant to alert users, and I don’t think users should be alerted when their battery is doing better than expected.

PR #43, opened 2015-07-22
Make pulseaudio dependency optional

lexszero:


stapelberg:
Why? i3status doesn’t have optional dependencies currently.

czarkoff:
This is a good case for optional dependency. PulseAudio is only one of audio daemons, and some may want to avoid having it.

stapelberg:
@czarkoff They don’t need to have it, they just need the library. We have other cases like that in i3status, like wireless information which doesn’t make much sense on typical desktop computers, yet requires libiw.

I’m going to close this as optional dependencies are not something we want to support in i3status.

rubin55:
Really? I think that is really a little bit to sharp through the corner; libiw is one thing where there essentially are no alternatives; pulseaudio is one of may optional audio management systems. Since you closed my ticket as duplicate, I'll repeat it here:

I noticed that 2.10 added PulseAudio support, which is great for those systems that run it, but an unneccesary dependency for those that don't have audio or that are running a minimal distribution.

I3 is a great gui for many types of uses and it would be nice if i3status would not inadvertently force a given audio subsystem; it would therefore be great if PulseAudio support could be optional.

stapelberg:
We don’t force you to actually use PulseAudio, we force you to have a (very likely deduplicated) libpulse in your memory. I still think that’s acceptable and the correct thing bottom line.

rubin55:
Well it's your software; have fun with it! I personally think you do actually force another layer of indirection onto any system that simply wants a few status items on their screen when they're using i3. In my case, I run a system with a very minimal set of dependencies which keeps my system very lean and mean (which is the way I like it). Having pulseaudio or its libraries on there, even though I'm not using it is precisely what I'm trying to avoid. Kind regards + On to the next wm for me I suppose!

Airblader:
i3status is not bundled with i3 and not required at all to run i3 or i3bar. If you want minimal dependencies, write a shell script that uses sysfs to generate the output for the status line. 

That said, we neither can nor want to force you to use i3. If you find that other tools serve your need better, go for it.

stapelberg:
> I personally think you do actually force another layer of indirection onto any system that simply wants a few status items on their screen when they're using i3

The i3status volume module conditionally selects a code path depending on how it is configured. If your system doesn’t use PulseAudio, you won’t be running the code that libpulse contains at all. Given that the i3status binary is not mlocked into memory, your kernel won’t even load libpulse’s pages into memory (or, if they were loaded proactively, they can be paged out). In conclusion, you’re not running libpulse and there’s no additional memory usage for libpulse.

If your “layer of indirection” refers to the single conditional `if` statement, then we just fundamentally disagree. If it doesn’t, I still don’t get what it is you’re referring to.

rubin55:
So it essentially is a build-time dependency?

stapelberg:
No, you can’t remove libpulse afterwards, because the dynamic linking loader needs to load libpulse, in case you want to actually use it. It’s a run-time dependency and depends on your config/system.

Out of curiosity, is libpulse0 actually absent from your system? The list of software which uses it is huge: https://cloud.githubusercontent.com/assets/55506/12080610/3cb0cdf0-b260-11e5-8e71-71ad4e1aeedd.png (Debian testing)

Also, which distribution are you using? If the distribution already has support for very fine-grained dependencies (like Gentoo’s use flags), perhaps a good way to solve the issue at hand would be a distribution-specific patch for i3status.

rubin55:
Yes, it's Gentoo indeed (so yes, really no libpulse here 8-). I'll write the maintainer of the Gentoo ebuild for i3status a msg to ask him to provide a patch for a pulseaudio use flag.

CuriousGentooman:
Here's a Gentoo-specific patch I whipped up to remove PulseAudio dependency:

http://pastebin.com/i2VnRmcK

Note that the ebuild doesn't call epatch_user, so you'll need to do add that to the src_prepare function, or just patch it manually. If you do it manually, do note that the patch is applied post the other preparations in the ebuild (i.e. the sed expression and removal of the *.1 man page), and will fail otherwise.

rubin55:
Great! I sent an email to the maintainer Kacper Kowalik referencing this issue and your patch.

czarkoff:
@stapelberg: so you encourage third-party patches to add optional dependency concept to i3status, but you don't allow such patches into repo? This policy has several negative effects on your project:

* Package maintainers have to maintain local patches, potentially redoing each others' work.
* Increased maintainership burden increases delivery time.
* Binaries in packages are not strictly based on your repo any more, so end users have hard time figuring out proper avenue for filing bugs.

Looks like net loss for the project.

rubin55:
I got a response from the maintainer: "Let's try to convince upstream first. This patch isn't something I'd like to maintain in Gentoo". How about it? Pretty please with sugar on top? :-)

stapelberg:
As I mentioned in https://github.com/i3/i3status/pull/43#issuecomment-128424495, we don’t want to support optional dependencies in i3status.

I suggested that Gentoo carries the patch because Gentoo users seem to be interested in reducing build dependencies. If the Gentoo community cares so much about this feature that they are willing to carry the maintenance burden, that’s fine with me.

l29ah:
How much should we pay for this feature to get included?

stapelberg:
i3status features aren’t for sale.

czarkoff:
@stapelberg It is not only Gentoo, but also all BSDs. Anyway, I would volunteer to split volume module into several backend-dependent modules and add infrastructure for selecting wanted modules and building dependencies list based on users' selection. Is that something you'd be willing to consider?

CuriousGentooman:
Gentleman, gentleman, let's not argue; looking through the code it's clear that having optional dependencies doesn't fit the style and philosophy of the package, and as such I'm inclined to agree with @stapelberg. And either way, as the maintainers it's their right to put out the package as they see fit.

With that said, Gentoo users do love to exercise our OCD in our setups, so I took it a step further and created an overlay that adds a PulseAudio USE flag, thereby making it an optional dependency. Here you go:

https://github.com/CuriousGentooman/i3status-opt-pa

netzverweigerer:
Hi.

@rubin55 You should consider simply abandoning i3status, not i3 in general. All i3status does is looking for information that will very likely differ between users and systems anyways, to add information you need to write ugly wrappers, and sane alternatives exist (I recommend "i3blocks"). I've written a good bunch of scripts that generate i3bar status command compatible output, too, in different scripting languages including Python and Bash and abandoned i3status entirely.

Also consider just building i3 from Git, somewhere in your home directory. I'm also questioning how many packages depend on pulseaudio and systemd nowadays and If that even makes sense anymore, but speaking of i3, I run fine here with Airblader's i3-gaps fork of i3 and i3blocks and it works for me.

Just my 0.02€: I don't think letting a small program that simply outputs text depend on anything makes sense at all, I would just look if the requirement to display an i3bar element are satisfied, and if not, display a text string that indicates that they're not. That's what I do with the scripts I wrote that basically do comparable stuff.

josephg5:
@stapelberg sounds like a dictator.. with the same attitude as the gnome/systemd superheroes? i reached here tracing some arguments about people slagging off i3. my £0.02 re this issue is to remove the pulseaudio dependency.. period! not optional either! 

>  If your system doesn’t use PulseAudio, you won’t be running the code that libpulse contains at all.
you shouldn't get any bugs, if you're not querying that absent library on a non-pulseaudio system. sadly i think i should look to move off i3 too.. if this is how the wind blows here :(

Airblader:
Please watch your tone.

PR #45, opened 2015-07-31
Add "mbox" and "maildir" modules for mail notification

czarkoff:
**Mbox** is pretty straight-forward tool to check difference between atime and mtime of given file. **Maildir** is slightly more complex - it looks for directory that appear to be maildirs and counts files in "new" subdirectory.

### DIFF
--- a/i3status.c
+++ b/i3status.c
@@ -410,9 +410,30 @@ int main(int argc, char *argv[]) {
         CFG_CUSTOM_MIN_WIDTH_OPT,
         CFG_END()};
 
+    cfg_opt_t mbox_opts[] = {
+        CFG_STR("prefix", "✉: ", CFGF_NONE),
+        CFG_STR("label", "default", CFGF_NONE),
+        CFG_CUSTOM_ALIGN_OPT,
+        CFG_CUSTOM_COLOR_OPTS,
+        CFG_CUSTOM_MIN_WIDTH_OPT,
+        CFG_END()};
+
+    cfg_opt_t maildir_opts[] = {
+        CFG_STR("prefix", "✉ ", CFGF_NONE),
+        CFG_STR("format", "%dir:%count", CFGF_NONE),
+        CFG_STR("separator", ", ", CFGF_NONE),
+        CFG_STR("no_mail", "", CFGF_NONE),
+        CFG_STR("exclude", "drafts spam trash", CFGF_NONE),
+        CFG_CUSTOM_ALIGN_OPT,
+        CFG_CUSTOM_COLOR_OPTS,
+        CFG_CUSTOM_MIN_WIDTH_OPT,
+        CFG_END()};
+
     cfg_opt_t opts[] = {
         CFG_STR_LIST("order", "{}", CFGF_NONE),
         CFG_SEC("general", general_opts, CFGF_NONE),
+        CFG_SEC("maildir", maildir_opts, CFGF_TITLE | CFGF_MULTI),
+        CFG_SEC("mbox", mbox_opts, CFGF_TITLE | CFGF_MULTI),
         CFG_SEC("run_watch", run_watch_opts, CFGF_TITLE | CFGF_MULTI),
         CFG_SEC("path_exists", path_exists_opts, CFGF_TITLE | CFGF_MULTI),
         CFG_SEC("wireless", wireless_opts, CFGF_TITLE | CFGF_MULTI),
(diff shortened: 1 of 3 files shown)


stapelberg:
I’m undecided whether having this in i3status is a good idea with regards to the philosophy of what we want to display and what we don’t want to display.

czarkoff:
Could you elaborate, please, - I can't find any document on this philosophical matter.

Anyway, I will definitely keep using and maintaining them as long as I use i3status, and I am perfectly fine with doing it locally. That said, lack of mail checker was a major roadblock between me and i3status before, and I suppose somebody else could find it useful as well - hense this pull request.

stapelberg:
See http://i3wm.org/i3status/manpage.html#_what_about_memory_usage_or_cpu_frequency

czarkoff:
I though that you were referring to something else, because mail notification is exactly a thing "*which you would look at occasionally anyways, like the current date/time, whether you are connected to a WiFi network or not, and if you have enough disk space to fit that 4.3 GiB download.*"

i3status.c
```diff
@@ -410,9 +410,30 @@ int main(int argc, char *argv[]) {
         CFG_CUSTOM_MIN_WIDTH_OPT,
         CFG_END()};
 
+    cfg_opt_t mbox_opts[] = {
+        CFG_STR("prefix", "✉: ", CFGF_NONE),
+        CFG_STR("label", "default", CFGF_NONE),
```
stapelberg: Would it be possible to use `NULL` instead of a magic word?

src/print_maildir_info.c
```diff
@@ -0,0 +1,287 @@
...
+    mailbox_t *mailbox;
+    char *p;
+
+    INSTANCE(path);
+
+    ignored = NULL;
+    mailboxes = NULL;
+    maildir = path;
+
+    setup_ignore_list(exclude);
```
stapelberg: Why is it necessary to create this list over and over again?

src/print_maildir_info.c
```diff
@@ -0,0 +1,287 @@
...
+
+    if (find_maildirs(path) == false) {
```
stapelberg: `if (!find_maildirs(path)) {`

src/print_maildir_info.c
```diff
@@ -0,0 +1,287 @@
...
+        }
+    }
+    (void)closedir(dir);
+
+    if (box->count == 0) {
+        free(box);
+        return true;
+    }
+
+    snprintf(box->path, PATH_MAX, "%s", dirname(path + strlen(maildir) + 1));
```
stapelberg: I think you should copy `path+strlen(maildir+1` before passing it to `dirname`. The manpage dirname(3) mentions:

> Both  dirname()  and  basename() may modify the contents of path, so it may be desirable to pass a copy when calling one of these functions.

Given that you actually use `path` later on, this might lead to subtle bugs on some platforms.

src/print_maildir_info.c
```diff
@@ -0,0 +1,287 @@
...
+#include <dirent.h>
+#include <libgen.h>
+#include <yajl/yajl_gen.h>
+#include <yajl/yajl_version.h>
+
+#include "i3status.h"
+
+#define MBNAMELEN 100
+
+#define do_or_die(x) \
```
stapelberg: I dislike hiding control flow behind this macro. I think it’d be clearer to the reader if we didn’t use this macro. Can you remove it please?

src/print_maildir_info.c
```diff
@@ -0,0 +1,287 @@
...
+#include <string.h>
+#include <sys/types.h>
+#include <dirent.h>
+#include <libgen.h>
+#include <yajl/yajl_gen.h>
+#include <yajl/yajl_version.h>
+
+#include "i3status.h"
+
+#define MBNAMELEN 100
```
stapelberg: This is unused, please remove.

src/print_maildir_info.c
```diff
@@ -0,0 +1,287 @@
...
+    size_t count;
+    mailbox_t *next;
+};
+
+char *maildir;
+mailbox_t *mailboxes;
+exclude_t *ignored;
+
+/*
+ * Free ignore list.
```
stapelberg: We actually have include/queue.h available in i3status, so can you please use the list macros which are defined there instead of re-inventing a linked list? :)

stapelberg:
That’s a fair point. Can you also update the manpage please?

src/print_maildir_info.c
czarkoff: Sorry, here it is
```
     char *
     dirname(const char *path);
```
And I was assuming it is the same elsewhere...

src/print_maildir_info.c
czarkoff: You want me to remove this one, or other explicit checks (`strcmp(s1, s2) == 0`, `p == NULL`, `p != NULL`, etc) as well?

stapelberg: Only the `== false`.

PR #46, opened 2015-07-31
Add missing fixes from OpenBSD ports tree

czarkoff:
Apparently most patches from OpenBSD ports tree already made their way into i3. These remain.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,6 +68,11 @@ CFLAGS += -idirafter yajl-fallback
 OBJS:=$(wildcard src/*.c *.c)
 OBJS:=$(OBJS:.c=.o)
 
+ifeq ($(OS),OpenBSD)
+OBJS:=$(filter-out src/pulse.o, $(OBJS))
+LIBS:=$(filter-out -lpulse, $(LIBS)) -lpthread
+endif
+
 src/%.o: src/%.c include/i3status.h
 	$(CC) $(CPPFLAGS) $(CFLAGS) -c -o $@ $<
 	@echo " CC $<"
--- a/Makefile
+++ b/Makefile
@@ -43,10 +43,6 @@ CFLAGS+=-I/usr/local/include/
 LDFLAGS+=-L/usr/local/lib/
 endif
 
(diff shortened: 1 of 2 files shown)


stapelberg:
Can you please use clang-format to format the code to make travis happy?

czarkoff:
Now it builds for me with no patchwork both on OpenBSD and Linux, and travis is happy.

czarkoff:
If you consider merging this pull request without collapsing commits, please notify me - I'd rather purge *Home Server* from history and rearrange diffs on my side.

Makefile
```diff
@@ -72,6 +68,11 @@ CFLAGS += -idirafter yajl-fallback
 OBJS:=$(wildcard src/*.c *.c)
 OBJS:=$(OBJS:.c=.o)
 
+ifeq ($(OS),OpenBSD)
+OBJS:=$(filter-out src/pulse.o, $(OBJS))
+LIBS:=$(filter-out -lpulse, $(LIBS)) -lpthread
```
stapelberg: Why are you compiling without PulseAudio on OpenBSD? PulseAudio is available and actively maintained (AFAICT) on OpenBSD: http://openports.se/audio/pulseaudio

czarkoff:
Sure, but it plays completely different role. On Linux PulseAudio manages audio devices. On OpenBSD it is just an interface on top of SndIO, and most programs speak to SndIO directly. I am not even sure that PulseAudio volume settings actually affect the system-wide volume on OpenBSD.

FWIW you might notice that most OpenBSD ports are compiled without pulseaudio support if it is optional.

Makefile
czarkoff: Sure, but it plays completely different role. On Linux PulseAudio manages audio devices. On OpenBSD it is just an interface on top of SndIO, and most programs speak to SndIO directly. I am not even sure that PulseAudio volume settings actually affect the system-wide volume on OpenBSD.

FWIW you might notice that most OpenBSD ports are compiled without pulseaudio support if it is optional.

stapelberg:
Can you squash your commits please?

PR #47, opened 2015-08-01
Use SndIO for volume module on OpenBSD

czarkoff:
Same code as in #46 with commits squashed into one.


czarkoff:
I need some more testing, apparently.

czarkoff:
Sorry, everything is fine here. Accidentally started building wrong branch.

PR #48, opened 2015-08-02
Add "mbox" module

czarkoff:
Separate pull request for `mbox` module. (Branched out from #45.)

### DIFF
--- a/i3status.c
+++ b/i3status.c
@@ -410,9 +410,20 @@ int main(int argc, char *argv[]) {
         CFG_CUSTOM_MIN_WIDTH_OPT,
         CFG_END()};
 
+    cfg_opt_t mbox_opts[] = {
+        CFG_STR("prefix", NULL, CFGF_NONE),
+        CFG_STR("postfix", NULL, CFGF_NONE),
+        CFG_STR("label", NULL, CFGF_NONE),
+        CFG_STR("label_no_mail", NULL, CFGF_NONE),
+        CFG_CUSTOM_ALIGN_OPT,
+        CFG_CUSTOM_COLOR_OPTS,
+        CFG_CUSTOM_MIN_WIDTH_OPT,
+        CFG_END()};
+
     cfg_opt_t opts[] = {
         CFG_STR_LIST("order", "{}", CFGF_NONE),
         CFG_SEC("general", general_opts, CFGF_NONE),
+        CFG_SEC("mbox", mbox_opts, CFGF_TITLE | CFGF_MULTI),
         CFG_SEC("run_watch", run_watch_opts, CFGF_TITLE | CFGF_MULTI),
         CFG_SEC("path_exists", path_exists_opts, CFGF_TITLE | CFGF_MULTI),
         CFG_SEC("wireless", wireless_opts, CFGF_TITLE | CFGF_MULTI),
@@ -576,6 +587,12 @@ int main(int argc, char *argv[]) {
             CASE_SEC("ipv6") {
                 SEC_OPEN_MAP("ipv6");
                 print_ipv6_info(json_gen, buffer, cfg_getstr(sec, "format_up"), cfg_getstr(sec, "format_down"));
+                SEC_CLOSE_MAP;
+            }
+
+            CASE_SEC_TITLE("mbox") {
+                SEC_OPEN_MAP("mbox");
+                print_mbox_info(json_gen, buffer, title, cfg_getstr(sec, "prefix"), cfg_getstr(sec, "postfix"), cfg_getstr(sec, "label"), cfg_getstr(sec, "label_no_mail"));
                 SEC_CLOSE_MAP;
             }
 
--- a/i3status.c
+++ b/i3status.c
@@ -411,10 +411,8 @@ int main(int argc, char *argv[]) {
         CFG_END()};
 
     cfg_opt_t mbox_opts[] = {
-        CFG_STR("prefix", NULL, CFGF_NONE),
-        CFG_STR("postfix", NULL, CFGF_NONE),
-        CFG_STR("label", NULL, CFGF_NONE),
-        CFG_STR("label_no_mail", NULL, CFGF_NONE),
+        CFG_STR("format", NULL, CFGF_NONE),
+        CFG_STR("format_no_mail", NULL, CFGF_NONE),
         CFG_CUSTOM_ALIGN_OPT,
         CFG_CUSTOM_COLOR_OPTS,
         CFG_CUSTOM_MIN_WIDTH_OPT,
@@ -592,7 +590,7 @@ int main(int argc, char *argv[]) {
 
             CASE_SEC_TITLE("mbox") {
                 SEC_OPEN_MAP("mbox");
-                print_mbox_info(json_gen, buffer, title, cfg_getstr(sec, "prefix"), cfg_getstr(sec, "postfix"), cfg_getstr(sec, "label"), cfg_getstr(sec, "label_no_mail"));
+                print_mbox_info(json_gen, buffer, title, cfg_getstr(sec, "format"), cfg_getstr(sec, "format_no_mail"));
                 SEC_CLOSE_MAP;
             }
 
--- a/i3status.c
+++ b/i3status.c
@@ -558,7 +558,7 @@ int main(int argc, char *argv[]) {
      * bytes — given that the output of i3status is used to display
      * information on screen, more than 1024 characters for the full line
      * (!), not individual plugins, seem very unlikely. */
-    char buffer[4096];
+    char buffer[BUFFER_SIZE];
 
     void **per_instance = calloc(cfg_size(cfg, "order"), sizeof(*per_instance));
     pthread_mutex_lock(&i3status_sleep_mutex);
--- a/i3status.c
+++ b/i3status.c
@@ -411,7 +411,7 @@ int main(int argc, char *argv[]) {
         CFG_END()};
 
     cfg_opt_t mbox_opts[] = {
-        CFG_STR("format", "✉: %mails", CFGF_NONE),
+        CFG_STR("format", "✉: %messages", CFGF_NONE),
         CFG_STR("format_no_mail", NULL, CFGF_NONE),
         CFG_CUSTOM_ALIGN_OPT,
         CFG_CUSTOM_COLOR_OPTS,
--- a/i3status.c
+++ b/i3status.c
@@ -411,7 +411,7 @@ int main(int argc, char *argv[]) {
         CFG_END()};
 
     cfg_opt_t mbox_opts[] = {
-        CFG_STR("format", NULL, CFGF_NONE),
+        CFG_STR("format", "✉: %mails", CFGF_NONE),
         CFG_STR("format_no_mail", NULL, CFGF_NONE),
         CFG_CUSTOM_ALIGN_OPT,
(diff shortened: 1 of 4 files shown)


man/i3status.man
```diff
@@ -48,6 +48,7 @@ general {
 }
 
 order += "ipv6"
+order += "mbox"
```
stapelberg: This should read mbox /var/mail/root, right?

man/i3status.man
```diff
@@ -208,6 +213,25 @@ best available public IPv6 address on your computer).
 
 *Example format_down*: +no IPv6+
 
+=== Mbox
+
+Verifies mail in a single UNIX mailbox file. Presentation is controled with
+four variables: "prefix", "label", "label_no_mail", and "postfix". Note, 
+"prefix" and "postfix" are concatinated with "label" (or "label_no_mail") with
```
stapelberg: It’d be more consistent with other i3status features if you could use `format` and `format_no_new_mail`, with %mails being replaced by the number of mails.

czarkoff: Currently there is no number to print - the module says either that there is mail or that there is no mail. I may rename "label" and "label_no_mail" to "format" and "format_no_mail" anyway, but I don't think that fiddling with access time, detecting mbox format and parsing it is worth the trouble.

stapelberg: As per https://en.wikipedia.org/wiki/Mbox, all parsing that’s necessary is just detecting whether a line starts with “From: ”, which I think is feasible.

czarkoff: This is one of four formats, actually.

src/print_mbox_info.c
```diff
@@ -0,0 +1,78 @@
...
+    if (exists && S_ISREG(sb.st_mode) && sb.st_size != 0 && sb.st_atime <= sb.st_mtime) {
+        /* Read the file and count the "From" header instances */
+        if ((f = fopen(path, "rw")) == NULL) {
+            START_COLOR("color_bad");
+            outwalk += sprintf(outwalk, "mbox: ");
+            (void)strerror_r(errno, outwalk, sizeof(buffer) - (outwalk - buffer));
+            outwalk = buffer + strnlen(buffer, sizeof(buffer));
+        }
+        while (fgets(s, LINELEN, f) != NULL)
+            if (strncmp("From:", s, 5) == 0)
```
stapelberg: wikipedia says the string is “From: ” (note the trailing whitespace). Can you use that instead? If not, why not?

Also, please replace 5 by `strlen("From: ")`

src/print_mbox_info.c
```diff
@@ -0,0 +1,78 @@
...
+                messages++;
+
+        /* If fgets(3) returned NULL due to some error, message count
+         * is likely wrong.  Print error instead of message count */
+        if (!feof(f)) {
+            (void)fclose(f);
```
stapelberg: This looks like an oversight: you’re closing f twice, here and after the if.

src/print_mbox_info.c
```diff
@@ -0,0 +1,78 @@
...
+            (void)strerror_r(errno, outwalk, sizeof(buffer) - (outwalk - buffer));
+            outwalk = buffer + strnlen(buffer, sizeof(buffer));
+        }
+        (void)fclose(f);
+
+        /* Restore access and modification times from stat(2) call
+         * and apply back to the file */
+        TIMESPEC_TO_TIMEVAL(&times[0], &(sb.st_atim))
+        TIMESPEC_TO_TIMEVAL(&times[1], &sb.st_mtim)
+        (void)utimes(path, times);
```
stapelberg: Won’t updating the atime/mtime signal to other programs that these messages have been read?

src/print_mbox_info.c
```diff
@@ -0,0 +1,78 @@
...
+
+        START_COLOR("color_degraded");
+        for (*outwalk = '\0'; (p = strstr(format, "%mails")) != NULL; format = p + 6) {
```
stapelberg: It’d be more consistent with the rest of the code if you could use a loop that resembles our other loops for handling the format string. See e.g. src/print_wireless_info.c

src/print_mbox_info.c
czarkoff: They were not read. The idea is that this notification stays until atime is newer then mtime, which indicates that their content was read.

src/print_mbox_info.c
czarkoff: Yes, this if was supposed to end with "goto out;".

src/print_mbox_info.c
czarkoff: There is only one format token to parse here, so IMO having single loop with faster iteration makes more sense then consistent double, which computes the length of token on every input character.

On the other hand, if you insist, I could switch to consistent behavior.

src/print_mbox_info.c
stapelberg: Ah, I see. Can you use the word “reset” in the comment, to make this clearer?

src/print_mbox_info.c
stapelberg: I agree that your method is more efficient, but I think there is value in having the code consistent. Can you please change it?

src/print_mbox_info.c
czarkoff: I'm not sure where you are getting this from.

1. There is "From " line (with space and no column), which is a first line of a message in mbox. It is quoted by some programs and is not by other programs. Depending on particular format of mbox, amount of "From " strings may (`mutt`, `mail`) or may not (Thunderbird) equal the amount of messages.
2. There is "From:" header (with column), which is  a header according to [RFC 2822](https://tools.ietf.org/html/rfc2822), and as such may be followed by any printable character. Per RFC 2822 only one "From:" header may appear in the message, but strings starting with "From:" are not quoted in message bodies.

I don't really care which one to pick, because none of them is reliable, so the resulting number will not necessarily show the number of messages.

Furthermore, if mbox actually contains read messages, the only way to reliably check the number of unread messages is to **parse** messages, detect headers and iterate through them. At this stage you already have a mail reader.

Now, given that predominant use cases of mboxes are `git am` and reports of cron jobs, I see no reason to bother with reading the file. As a proof to my words, most mailcheckers (save for `biff`, which just prints the first *n* lines of mbox) don't do that.

stapelberg: Ah, yeah, I meant “From ”. I think we should use “From ”, as that seems to be the more widely supported one.

src/print_mbox_info.c
czarkoff: Done.

czarkoff:
Well, this version works for me with `mutt` and `mail`. And it even caches enough data to avoid parsing the mbox every time.

src/print_mbox_info.c
```diff
@@ -0,0 +1,140 @@
...
+ * When done, restore mtime and atime attributes of mbox, so that it remains
+ * visible on next check.
+ *
+ */
+void print_mbox_info(yajl_gen json_gen, char *buffer, const char *path, const char *format, const char *format_no_mail) {
+    char *outwalk = buffer;
+    const char *walk = format;
+    struct stat sb;
+    bool exists = !stat(path, &sb);
+    bool after_newline = true, in_header = false;
```
stapelberg: I think a better name for `after_newline` is `after_blank_line`, since every line is terminated with a newline characters, but that’s not what you mean :).

src/print_mbox_info.c
```diff
@@ -0,0 +1,140 @@
...
+
+#include "i3status.h"
+
+#define LINELEN 1000
+
+/* These variables are used to store data between calls to print_mbox_info().
+ * The purpose of such */
+struct timeval times[2];
+size_t messages = 0;
+bool initialized = false;
```
stapelberg: Instead of using `initialized`, why not define `struct timeval times[2] = { 0 };`, which has the same effect?

src/print_mbox_info.c
```diff
@@ -0,0 +1,140 @@
...
+#include <yajl/yajl_version.h>
+
+#include "i3status.h"
+
+#define LINELEN 1000
+
+/* These variables are used to store data between calls to print_mbox_info().
+ * The purpose of such */
+struct timeval times[2];
+size_t messages = 0;
```
stapelberg: Can you please mark `times` and `messages` as static, since they are only used within this source file?

src/print_mbox_info.c
```diff
@@ -0,0 +1,140 @@
...
+    char s[LINELEN];
+
+    INSTANCE(path);
+
+    if (!initialized) {
+        times[1].tv_sec = 0;
+        initialized = true;
+    }
+
+    if ((exists && S_ISREG(sb.st_mode)) || !exists) {
```
stapelberg: I find the control flow rather hard to read. Can you refactor it in such a way that you check for error conditions early on and jump to `out`? E.g.:

```c
const bool unmodified = sb.st_mtim.tv_sec == times[1].tv_sec;
if (!exists || sb.st_size == 0 || (unmodified && messages == 0)) {
  START_COLOR("color_good");
  outwalk += sprintf(outwalk, "%s", format_no_mail);
  END_COLOR;
  goto out;
}

// … (can assume exists is true now)
```

czarkoff:
In last commit I also added several new tokens and renames "%mails" to "%messages". I somehow forgot to mention that in the commit message. Anyway, all of these commits (except for 1a05f2a9a2b27f7369cc6bd4bf71a65bbe6406d9) should be squashed before merging.

man/i3status.man
```diff
@@ -208,6 +213,25 @@ best available public IPv6 address on your computer).
 
 *Example format_down*: +no IPv6+
 
+=== Mbox
+
+Verifies mail in a single UNIX mailbox file. Presentation is controled with
+two variables: "format" and "format_no_mail". Keywords "%messages", "%name" and
+"%path" in format strings would be expanded into message count, last element of
```
stapelberg: Why are %path and %name useful? They only depend on the configuration file, not on any information that is gathered at runtime, so users might as well configure them statically, right?

man/i3status.man
```diff
@@ -208,6 +213,25 @@ best available public IPv6 address on your computer).
...
+path and whole path respectively.
+
+The mbox implementation in this module expects "mboxo" mboxes, so it may report
+wrong numbers for other types of mboxes.  In particular, it doesn't check
+"Content-Length" header.
+
+*Example configuration*:
+-------------------------------------------------------------
+mail "/var/mail/user" {
+    format = "%mails new messages for %name"
```
stapelberg: This says %mails, but should say %messages

i3status.c
```diff
@@ -410,9 +410,18 @@ int main(int argc, char *argv[]) {
         CFG_CUSTOM_MIN_WIDTH_OPT,
         CFG_END()};
 
+    cfg_opt_t mbox_opts[] = {
+        CFG_STR("format", "✉: %mails", CFGF_NONE),
```
stapelberg: This says %mails but should say %messages

man/i3status.man
czarkoff: One might want to share i3status.conf between several users.Or, if one watches for several mboxes, he might want to be able to tune string for one of them and copy-paste it.

stapelberg: These are not particularly strong use-cases (as they are trivially fulfilled with a text editor) and we don’t support them for other modules. Can you please remove the format strings?

PR #49, opened 2015-08-06
New "mail" module

czarkoff:
This request superceeds #45 and #48. This time it is a single module, that, unlike modules in those two previous requests, may be used several times in one i3status configuration without screwing the report. I tried to keep code documented, so that it should be easy to comprehend. I also tried to avoid wasting excessive amounts of memory and keep module safe from potential memory corruption issues (some people have really monstrous mailbox structure).

### DIFF
--- a/i3status.c
+++ b/i3status.c
@@ -410,9 +410,22 @@ int main(int argc, char *argv[]) {
         CFG_CUSTOM_MIN_WIDTH_OPT,
         CFG_END()};
 
+    cfg_opt_t mail_opts[] = {
+        CFG_STR("mailpath", NULL, CFGF_NONE),
+        CFG_STR("ignore", NULL, CFGF_NONE),
+        CFG_INT("width", MAIL_REPORT_WIDTH, CFGF_NONE),
+        CFG_STR("format", "✉: {%mailbox:%messages}", CFGF_NONE),
+        CFG_STR("format_no_mail", NULL, CFGF_NONE),
+        CFG_STR("replacement", " …", CFGF_NONE),
+        CFG_CUSTOM_ALIGN_OPT,
+        CFG_CUSTOM_COLOR_OPTS,
+        CFG_CUSTOM_MIN_WIDTH_OPT,
+        CFG_END()};
+
     cfg_opt_t opts[] = {
         CFG_STR_LIST("order", "{}", CFGF_NONE),
         CFG_SEC("general", general_opts, CFGF_NONE),
+        CFG_SEC("mail", mail_opts, CFGF_TITLE | CFGF_MULTI),
         CFG_SEC("run_watch", run_watch_opts, CFGF_TITLE | CFGF_MULTI),
         CFG_SEC("path_exists", path_exists_opts, CFGF_TITLE | CFGF_MULTI),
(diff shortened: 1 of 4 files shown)


czarkoff:
It seems I've finished tinkering with history, so this pull request can be reviewed now. I'll collapse all changes once you are ready to merge.

stapelberg:
I see that you put in a lot of effort into this, but I’m starting to doubt whether having this module in i3status itself is a good idea. The fact that inventing a custom format strings seems to be necessary to cater to basic use-cases is not a good sign.

Maybe this functionality should instead go into a script in contrib/. That way, you could also use existing modules to parse maildir/mbox and we could avoid that code duplication. Also, you’d have more freedom with regards to config formats and how often you’d want to break them to support new use-cases.

The proposed module has grown to be one of the largest ones that i3status would have — the only other modules this big are battery and wifi, both of which have implementions for 3 different operating systems.

What do you think?

czarkoff:
>The fact that inventing a custom format strings seems to be necessary to cater to basic use-cases is not a good sign.
> […]
> The proposed module has grown to be one of the largest ones that i3status would have

Custom format string is there because the module is supposed to present a lot of similar data pieces, so consistency in their presentation makes sense. This module could be reorganized to present single mailbox per *order* entry. This would allow dropping one level of singly-linked list and reducing the code by up to ¼ at expense of loosing prefixes and postfixes, forcing users to duplicate configuration entries and (probably) removing total message count. I guess almost ⅓ could be dropped with checking for names maildirs instead of huge maildir++. In other words, the complexity of this module may be reduced, if SLOC count and format strings syntax consistency are that important. That was not my goal though, and that's why the module is what it is.

FWIW if I was making *ethernet* and *wireless* modules, I would also make them one module and use similar format string with default format like "*N {%interface: %ip %essid signal}*". There is nothing inherently different about mail checking.

> Maybe this functionality should instead go into a script in contrib/.

As I gather, that makes sense for things that require custom update interval, which is definitely not the case here. Furthermore, this would lift limitation of periodic polling, so that the code (at least in my hands :smile:) will turn into a kqueue/epoll jungle. Even I am scared of this perspective.

On the other hand, doing the same in python would reduce this module to 100 SLOC or so, I guess. At least the checking part would be almost a one-liner. I am not sure about performance implications though – python script will likely bog down the whole setup under more or less heavy disk load, while this module performs well with 1 second interval even under heavy IO activity.

> That way, you could also use existing modules to parse maildir/mbox and we could avoid that code duplication.

That would be a nightmare. Most of what I see depends on glib, qt or some other monster lib I'd rather not venture into.

> What do you think?

My thoughs on the subject are expressed in commits above. I rely on mail notification in my status bar, and currently there is no way for me to get it without such module.

I could draft another version with less complexity (as described above), and a python script for `contrib/`, if you think that either is a good idea.

stapelberg:
I agree with your points. I think adding a script in python, or whichever other high-level language you like that suits the task at hand, to the contrib/ folder would be a good idea. In case there is lots of user demand for the feature to be included with i3status itself, we can always revisit that with the new data and insights we get from providing a module in contrib/.

czarkoff:
OK, closing this pull request for now. I'll open new pull request once python version is available.

PR #53, opened 2015-08-24
Add IBus status display

puiterwijk:
Signed-off-by: Patrick Uiterwijk <puiterwijk@redhat.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ addons:
     - libyajl-dev
     - libasound2-dev
     - libpulse-dev
+    - libibus-1.0-dev
     - libiw-dev
     - asciidoc
     # xmlto pulls in a2x which we need for generating manpages from asciidoc.
--- a/Makefile
+++ b/Makefile
@@ -47,6 +47,17 @@ ifeq ($(OS),NetBSD)
 LIBS+=-lprop
 endif
 
+ifeq ($(OS),Linux)
+CFLAGS+=-pthread
+CFLAGS+=-I/usr/include/ibus-1.0
+CFLAGS+=-I/usr/include/glib-2.0
+CFLAGS+=-I/usr/lib64/glib-2.0/include
+LIBS+=-libus-1.0
+LIBS+=-lgio-2.0
+LIBS+=-lgobject-2.0
+LIBS+=-lglib-2.0
+endif
+
 # This probably applies for any pkgsrc based system
 ifneq (, $(filter $(OS), NetBSD DragonFly))
 CFLAGS+=-I/usr/pkg/include/
@@ -67,6 +78,10 @@ CFLAGS += -idirafter yajl-fallback
 
 OBJS:=$(wildcard src/*.c *.c)
 OBJS:=$(OBJS:.c=.o)
+
+ifneq ($(OS),Linux)
+OBJS:=$(filter-out src/print-ibus.o, $(OBJS))
+endif
 
 ifeq ($(OS),OpenBSD)
 OBJS:=$(filter-out src/pulse.o, $(OBJS))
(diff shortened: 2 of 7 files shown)


stapelberg:
Can you outline your use-case please? AFAIK, ibus provides a tray icon, why is that not enough to see the current status? Also, why is it important to see the status at all? Whenever I deal with ibus, I only ever set the mode I want to use and leave it at that.

puiterwijk:
I regularly switch ibus engines between Dutch, English (US) and kkc, and would like to be able to see in textual form what the current state is.
A textual form is in my opinion easier to interpret quickly vs the small icon displayed in i3bar.

puiterwijk:
Also, regarding your comment in #34: that on the first keypress you will see whether you have the correct layout selected, that has some limitations:
- If you are entering a password, you may enter it incorrectly multiple times before trying to switch keyboard layout
- If you are using a program mainly controlled by keyboard, you might have initiated an action before noticing you had the wrong layout selected (like having selected "n" to a question while you wanted another response)

True, both are also still possible with a textual display, but it is possible to train oneself to look at the status before entering a password or executing a command.

stapelberg:
Please comment on #34 directly if you have comments about #34, it’s not good to mix up discussions.

> I regularly switch ibus engines between Dutch, English (US) and kkc, and would like to be able to see in textual form what the current state is.

I tried answering this question first by reading https://github.com/ibus/ibus/wiki, but that page is not helpful in my situation. What does ibus actually do, compared to e.g. switching your XKB group from one keyboard layout to the next when loading multiple keyboard layouts? Is it roughly the same thing, comparable, or something else entirely?

Is the ibus engine the only thing that influences your input, or might other people need to display different information from ibus?

puiterwijk:
ibus is one way to handle keyboard layouts and Input Method Editors (IMEs, ways to input CJK languages).
For setups where people use ibus, it is the only thing influencing the keyboard layout in applications configured to use it.

For example GNOME-shell (GNOME 3) uses only ibus to change keyboard mappings.

While it is possible for users to make applications ignore ibus when starting them by (not) setting the required environment variables, I think that most people that use ibus just have it in their profile config, and as such use it for all applications.

stapelberg:
Thanks for the explanation, but it’s still a bit too high-level for me. How does ibus actually work? Where does it hook into? Does it grab keys? How does ibus “influence”/“handle” the keyboard layout?

puiterwijk:
IBus is implemented in the form of input module drivers for GTK/QT/X.
So one would set environment variables (for example XMODIFIERS=@im=ibus), to make it use ibus instead of the X Input Method (XIM).

So whenever you press any keys in applications that use it, it will send the keypresses to its daemon (in the case ibus), which will then send it on to an input engine.

The input engine is what handles the specific input method, there are a lot of engines like English US, Dutch, Anthy (one of the Japanese ones), etc.

For extended latin character sets, the engine will pretty much directly pass the pressed key back to ibus which passes it back to the application (except for dead keys, in which case it might wait if you pressed for example the quote to allow you to enter accented characters).
In non-Latin character sets (CJK languages mostly), or with "typing booster" engines (for example ispell which provides autocompletion just like on smart phones), it will start a pre-edit context, and pass the information and the current candidates (possible meanings of the entered text) on to the application, so that a UI can be displayed that allows the user to select the text they meant.
After the user presses enter, the final characters are provided to the application through the IM Module hook points.

stapelberg:
Thanks for the explanation, the crucial part for me was that IBus uses hooks in the toolkits instead of influencing the X11 server layout.

I agree that it might make sense to display the current IBus state for people who do actively use IBus.

Given that you are the first person to ever mention IBus in the context of i3status to me, I’m skeptical that adding it to i3status itself makes sense, though. We want to keep i3status small, and we can’t just add any piece of system information to it if we want to succeed with that goal.

Could you instead implement this as a wrapper around i3status and place it in contrib/? See http://i3wm.org/i3status/manpage.html#_external_scripts_programs_with_i3status

Airblader:
Closing since there has been no activity and implementing it in a wrapper script was proposed.

PR #60, opened 2015-09-15
Allow pulseaudio to be disabled at build time.

bsdjhb:
- If DISABLE_PULSEAUDIO is defined as a make variable, then disable
  use of pulseaudio.  This allows packages to optionally include
  pulseaudio support (for example in FreeBSD ports pulseaudio could
  become an OPTION on the next update).
- Explicitly link in -lpthread always.  The main loop uses pthread
  symbols directly, and on some systems libc only provides non-functional
  stubs.  On FreeBSD the pthread_cond_timedwait() stub returns instantly
  resulting in i3status never sleeping.
- Define DISABLE_PULSEAUDIO on OpenBSD by default to preserve existing
  behavior.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,7 @@ CFLAGS+=-Iinclude
 LIBS+=-lconfuse
 LIBS+=-lyajl
 LIBS+=-lpulse
+LIBS+=-lpthread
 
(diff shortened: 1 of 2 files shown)


stapelberg:
duplicate of #43

PR #66, opened 2015-10-01
Implemented Custom Display Section

JSmith-BitFlipper:
What I noticed from i3status when I first began configuring it was that it was not very flexible in configuration. It did not support any add-ons or any custom sections to be displayed or the ability to format the output of a given section differently.

I decided the past few days to implement this extendability. Basically, I created the ability to add a "custom" section in the i3status.conf file. Within this section is a parameter that takes a string of inline bash code or a bash file to execute. It then executes that code or file and its stdout output is recorded and displayed as its own area in i3status. This allows i3status users to write custom bash scripts and present the relative information in the i3status bar. This allows for the creation of pseudo-add-ons such as a "new email checker" for i3status.

Sample i3status.conf custom section:
```
...
order += "custom some-title"
order += "custom some-other-title"
...

custom some-title {
        exec = "printf \"This is a sample inline code exec example\";"
}

custom some-other-title {
        exec = "/home/<user>/full/path/to/bash/script/to/execute.sh"
}
```

Since some scripts may take a bit of time to execute, all custom sections run in their own separate threads in an infinite loop. After a script executes, it then waits for an unblock signal that comes from the main i3status loop once per iteration (more specifically, the unblock signal is signaled to a given custom's thread after every time the output buffer of that custom's thread is printed). That way, scripts that execute quickly will not needlessly execute many times in rapid succession and will only execute aligned to the iterations of the main i3status loop. This is done using pthreads.

Note:
As this code was written on a Linux machine, I was only able to test it on Linux. I noticed that i3status has some cross-platform comparability macros for other operating systems. In particular for threading, I am not sure of its cross-platform compatibility. Therefore, if it is not compatible, a simple "Linux-only" macro can be created as a temporary fix to permit custom sections to run only on Linux.


stapelberg:
From http://i3wm.org/i3status/manpage.html#_external_scripts_programs_with_i3status

> In i3status, we don’t want to implement process management again. Therefore, there is no module to run arbitrary scripts or commands. Instead, you should use your shell, for example like this:

Sorry for the effort that you put into this pull request.

Airblader:
Regarding Michael's last sentence we should also state that we recommend bringing features up for discussion first to see if they'd be accepted at all and also to shape the form they should take.

JSmith-BitFlipper:
I would understand why such functionality is avoided if running arbitrary scripts presents any security issues or causes anything undesirable to i3status. However, may I still ask why "we don’t want to implement process management again" and why a module that runs arbitrary scripts or commands is disregarded?

Airblader:
i3status is aimed to be lightweight and invoking a shell all the time really breaks that. Given that any user can still pre- or append arbitrary things as described in the man page Michael referenced, the benefit of adding this into i3status doesn't outweigh the disadvantages and maintenance effort. In fact I believe Michael stated that in retrospect there's modules in i3status he wouldn't want to be there, but we have a strong aim on not breaking configs.

If you want a more extensive and flexible output for i3bar that also executes shell scripts I would generally suggest to move to an alternative such as i3blocks.

It's simply not i3status' aim to implement everything that can be implemented. For example, we intentionally don't support all the properties the i3bar protocol offers.

PR #68, opened 2015-10-03
Added support for lemonbar

bsvh:
This adds support for lemonbar (https://github.com/LemonBoy/bar).

### DIFF
--- a/i3status.c
+++ b/i3status.c
@@ -163,11 +163,21 @@ static int parse_min_width(cfg_t *context, cfg_opt_t *option, const char *value,
  *
  */
 static int valid_color(const char *value) {
-    if (strlen(value) != 7)
-        return 0;
+    const int len = strlen(value);
+
+    if (output_format == O_LEMONBAR) {
+        /* lemonbar supports an optional alpha channel */
+        if (len != strlen("#rrggbb") && len != strlen("#aarrggbb")) {
+            return 0;
+        }
+    } else {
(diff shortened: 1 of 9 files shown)


i3status.c
```diff
@@ -163,11 +163,18 @@ static int parse_min_width(cfg_t *context, cfg_opt_t *option, const char *value,
...
-        return 0;
+    int len;
+
+    /* lemonbar supports an optional alpha channel */
+    if ((len = strlen(value)) != 7) {
+        if (output_format != O_LEMONBAR)
+            return 0;
+        else if (len != 9)
+            return 0;
+    }
```
stapelberg: I think the following code makes it a bit easier to understand the logic:
```c
const int len = strlen(value);
if (output_format == O_LEMONBAR) {
  /* lemonbar supports an optional alpha channel */
  if (len != strlen("#rrggbb") && len != strlen("#rrggbbaa")) {
    return 0;
  }
} else {
  if (len != strlen("#rrggbb")) {
    return 0;
  }
}
```

stapelberg:
Looks good overall.

bsvh:
Ok, I updated the color validation like you asked.

stapelberg:
Can you please use clang-format to format your code?

Airblader:
And maybe also squash those commits. ;)

bsvh:
Formatted and squashed.

PR #69, opened 2015-10-03
Adding new tube status script

ivernus:
Adding a script to eh contrib dir. I made this for my own i3 to keep an eye on the TFL tube status. I tried to get it working with the wrapper script but couldn't manage it.


stapelberg:
AFAICT, this script sends one request to api.tfl.gov.uk every `interval` (defaulting to 5s but often changed to 1s). I think that’s a bit too much. Could you please change the script to be less aggressive?

brioscaibriste:
Thanks Michael, good point, I will add a time delay option and hopefully re request merge soon.

Airblader:
You don't need to close the pull request, just force push onto the branch and Github will update it. We can wait. :9

brioscaibriste:
Thanks for the tip Airblader :)

brioscaibriste:
Hi folks, just dropped a quick update of the scripts with some throttling in place. It can be tuned to run every X minutes, otherwise caching the line status data in /tmp. I hope this is correct and makes a bit more sense than the original one.

contrib/tubestatusdef.py
```diff
@@ -0,0 +1,167 @@
...
+ two functions at least, one for time, one for polling, returning status value
+'''
+
+def Throttle(PollInterval,Throttle,TFileName):
+
+    # Does the temporary file exist
+    TFileStatus = os.path.isfile(TFileName)
+    
+    # Current epoch time
+    CurrentStamp = str(time.time()).split('.')[0]
```
stapelberg: Use `CurrentStamp = int(time.time())` instead.

contrib/tubestatusdef.py
```diff
@@ -0,0 +1,167 @@
...
+
+ check if temp file exists, if not, write it with time stamp in it, else pull time stamp
+ check time stamp vs. poll interval, if it's less than time + interval, skip it, else run and write the temp file again
+ two functions at least, one for time, one for polling, returning status value
+'''
+
+def Throttle(PollInterval,Throttle,TFileName):
+
+    # Does the temporary file exist
+    TFileStatus = os.path.isfile(TFileName)
```
stapelberg: Why introduce a variable that you only use once? You can use `if os.path.isfile(TFileName)` below, which is more clear when reading.

contrib/tubestatusdef.py
```diff
@@ -0,0 +1,167 @@
...
+        TFile.closed
+ 
+        # Set the Remainder high to force the next run
+        Remainder = 1000000 
+
+    # The poll interval in seconds
+    PollInterval = PollInterval * 60
+
+    # If the remainder is less than the poll interval don't run the command, if it isn't run the command
+    if ( Remainder < PollInterval ):
```
stapelberg: More idiomatic:

```python
if Remainder < PollInterval:
  return False
# …
return True
```

contrib/tubestatusdef.py
```diff
@@ -0,0 +1,167 @@
...
+
+def RetrieveTFLData(Line,Run,SFileName):
+
+    # TFL Unified API URL
+    TFLDataURL = "https://api.tfl.gov.uk/Line/" + Line + ("/Status?detail=False"
+    "&app_id=&app_key=")
+
+    if Run: 
+        # Read all the information from JSON at the specified URL
+        RawData = urlopen(TFLDataURL).readall().decode('utf8') or die("Error, failed to "
```
stapelberg: Are you aware of http://docs.python-requests.org/en/latest/? That module might make the code here more readable/shorter.

contrib/tubestatusdef.py
```diff
@@ -0,0 +1,167 @@
...
+        TFLData = json.loads(RawData)
+
+        # Sanitize the data to get the line status
+        Scratch = (TFLData[0]['lineStatuses'])
+        LineStatusData = (Scratch[0]['statusSeverityDescription'])
+        
+        # Cache the staus in a file
+        with open(SFileName, 'w+') as SFile:
+            SFile.write(LineStatusData)
+        SFile.closed
```
stapelberg: Here and below, you don’t need to call `.closed` — when using `with`, the file is automatically closed as soon as you leave the scope of the `with` block. The python docs merely call `.closed` to illustrate that the file is indeed closed. In your code, it doesn’t have any effect.

contrib/tubestatusdef.py
brioscaibriste: I've tried to change this to using int rather than strings but can't quite get it to work as I would like.

stapelberg: Can you give me more details about what doesn’t work (i.e. the code and the error message you get)? Without anything else, I can’t really answer this.

Airblader:
@brioscaibriste @stapelberg What's the status here?

brioscaibriste:
Hi all

I sent a few mails to Transport For London with regards to using my API key
in open source software. They still haven't replied. I will continue to
chase.

Thanks

Conor

Airblader:
@ivernus So… what now? This PR has been open for a long, long time now.

ivernus:
@Airblader I'm more than happy for it to be merged, it was paused for a bit as TFL would never confirm what to do if you wanted to have a key for open source software. I put in the documentation that an API key is required so it should be good to merge.

Airblader:
Thanks for the answer. There's still a review comment you haven't addressed. Also before we can merge it you'll need to squash the commits so we have a clean git log. But before you do so @stapelberg should perhaps say what his thoughts on it are now. :)

stapelberg:
Given how big the source of this contrib script is, I wonder whether it’ll be easier if it was maintained in a third-party repository. That way changes could be made more quickly and you’re not tied to the i3status release cycle.

Thoughts? I.e., is there any specific reason why this should be bundled with i3status?

ivernus:
that would work @stapelberg , I made it specifically for i3status really as I use that most of the time, so I thought it would be good bundled in. But I'm happy to just keep it separate if you think it would be better.

stapelberg:
Thanks for the update.

I think it might be best to put it into a repository owned by you, possibly called tubestatus-for-i3status (like there is quickswitch-for-i3, for example).

I’ll close the PR.

Issue #31, opened 2015-04-30
Pango markup setting


ixjlyons:
I see that Pango markup support was [added to `i3bar`](https://github.com/i3/i3/commit/e681f34ec1af4b19f9dfd9c0e8d9c29beaf3e2db) and quickly [made non-default](https://github.com/i3/i3/commit/beaf90494300640e3f811b3b99bb84296cd716e7). When it was default, it was possible to use markup in the `i3status` config -- for example:

````
disk "/" {
    format = "<span size='large' color='#ffffff'>disk:</span> %avail
}
```

I'm wondering if there would be any interest in adding a `markup` option to `i3status` general config. This would allow users who know about the markup (and consequently know of the need to escape ampersands, angle brackets, etc.) to use it, but those who don't care about additional formatting would be unaffected.

I realize, however, that features which are achievable some way other than changing the `i3status` source are generally not added, and this falls into that category. I wrote a simple wrapper script to inject `"markup": "pango"` into the JSON before passing it along to `i3bar`, and it seems to work fine. I also took my first look at the source and hacked together a solution there. It was fairly straightforward to add, so I wanted to see what others thought about it.

stapelberg:
I think in general it’d be okay to have a `markup` option in i3status. Note that all output not directly hard-coded in i3status or supplied by the user would need to be escaped for pango markup. This might be relevant for the time module (I haven’t checked all timezones), and is definitely relevant for the wireless module which displays the ESSID.

I suppose we’d need to re-implement http://sources.debian.net/src/glib2.0/2.44.0-2/glib/gmarkup.c/?hl=2227#L2149 for that.

Unless there is something which I haven’t thought of, I think I’d merge a pull request which adds a markup option.

ixjlyons:
Of course I hadn't thought of automatically generated content. I need to have a bit more of a look at the `i3status` source, but I'm still interested in this as a feature.

@stapelberg: how do you recommend accessing the config from 	1kj

stapelberg:
> how do you recommend accessing the config from 1kj

What’s “1kj”? I don’t follow.

ixjlyons:
Whoops, sorry about that. I hadn't completed my thought but the comment somehow got submitted (my cat is looking suspect:).

Anyway, here's more or less what I was going to ask:

[...]

PR #72, opened 2015-10-09
Pango setting

ixjlyons:
Implements #31.

I think this is more or less ready to go save the following:

* The added documentation is...minimal at best. Of course it's not necessary to reproduce the Pango markup docs, but it might be nice to have an example.

* Only printing the wireless SSID has been updated. Is there any other generated content that might need escaping (i.e. containing `&`, `<`, `>`, `'`, `"`)?

Finally, I haven't squashed the commits because I'm assuming there will be comments :)

### DIFF
--- a/include/i3status.h
+++ b/include/i3status.h
@@ -8,6 +8,9 @@ enum { O_DZEN2,
        O_TERM,
        O_NONE } output_format;
 
+enum { M_PANGO,
+       M_NONE } markup_format;
+
 char *pct_mark;
 
 #include <stdbool.h>
@@ -78,6 +81,9 @@ char *pct_mark;
          * not forgotten in the module */                                                        \
         *outwalk = '\0';                                                                         \
         if (output_format == O_I3BAR) {                                                          \
+            char *_markup = cfg_getstr(cfg_general, "markup");                                   \
+            yajl_gen_string(json_gen, (const unsigned char *) "markup", strlen("markup"));       \
+            yajl_gen_string(json_gen, (const unsigned char *)_markup, strlen(_markup));          \
             yajl_gen_string(json_gen, (const unsigned char *) "full_text", strlen("full_text")); \
             yajl_gen_string(json_gen, (const unsigned char *)text, strlen(text));                \
         } else {                                                                                 \
@@ -176,6 +182,7 @@ void print_separator(const char *separator);
 char *color(const char *colorstr);
 char *endcolor() __attribute__((pure));
 void reset_cursor(void);
+void maybe_escape_markup(char *text, char **buffer);
 
 /* src/auto_detect_format.c */
 char *auto_detect_format();
@@ -193,7 +200,7 @@ const char *first_eth_interface(const net_type_t type);
 void print_ipv6_info(yajl_gen json_gen, char *buffer, const char *format_up, const char *format_down);
 void print_disk_info(yajl_gen json_gen, char *buffer, const char *path, const char *format, const char *format_not_mounted, const char *prefix_type, const char *threshold_type, const double low_threshold);
 void print_battery_info(yajl_gen json_gen, char *buffer, int number, const char *path, const char *format, const char *format_down, const char *status_chr, const char *status_bat, const char *status_full, int low_threshold, char *threshold_type, bool last_full_capacity, bool integer_battery_capacity, bool hide_seconds);
-void print_time(yajl_gen json_gen, char *buffer, const char *title, const char *format, const char *tz, time_t t);
+void print_time(yajl_gen json_gen, char *buffer, const char *title, const char *format, const char *tz, const char *format_time, time_t t);
 void print_ddate(yajl_gen json_gen, char *buffer, const char *format, time_t t);
 const char *get_ip_addr(const char *interface);
 void print_wireless_info(yajl_gen json_gen, char *buffer, const char *interface, const char *format_up, const char *format_down);
--- a/src/output.c
+++ b/src/output.c
@@ -78,3 +78,46 @@ void print_separator(const char *separator) {
 void reset_cursor(void) {
     printf("\033[?25h");
 }
+
+/*
+ * Escapes ampersand, less-than, greater-than, single-quote, and double-quote
+ * characters with the corresponding Pango markup strings if markup is enabled.
+ * See the glib implementation:
+ * https://git.gnome.org/browse/glib/tree/glib/gmarkup.c?id=03db1f455b4265654e237d2ad55464b4113cba8a#n2142
+ *
+ */
+void maybe_escape_markup(char *text, char **buffer) {
+    if (markup_format == M_NONE) {
+        *buffer += sprintf(*buffer, "%s", text);
+        return;
+    }
+    for (; *text != '\0'; text++) {
+        switch (*text) {
+            case '&':
+                *buffer += sprintf(*buffer, "%s", "&amp;");
+                break;
+            case '<':
+                *buffer += sprintf(*buffer, "%s", "&lt;");
+                break;
+            case '>':
+                *buffer += sprintf(*buffer, "%s", "&gt;");
+                break;
+            case '\'':
+                *buffer += sprintf(*buffer, "%s", "&apos;");
+                break;
+            case '"':
+                *buffer += sprintf(*buffer, "%s", "&quot;");
+                break;
+            default:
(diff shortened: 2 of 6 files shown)


include/i3status.h
```diff
@@ -71,10 +74,15 @@ enum { O_DZEN2,
  * format is JSON) or just output to stdout (any other output format). */
 #define OUTPUT_FULL_TEXT(text)                                                                   \
     do {                                                                                         \
-        /* Terminate the output buffer here in any case, so that it’s                          \
+        /* Terminate the output buffer here in any case, so that it’s                            \
          * not forgotten in the module */                                                        \
         *outwalk = '\0';                                                                         \
         if (output_format == O_I3BAR) {                                                          \
+            char *_markup = cfg_getstr(cfg_general, "markup");                                   \
+            if (_markup) {                                                                       \
```
stapelberg: I don’t think the conditional is necessary, is it? cfg_getstr’s docs say the default value will be returned if the option was not set, and you declare the default to be `"none"`.

src/output.c
```diff
@@ -73,3 +73,39 @@ void print_separator(const char *separator) {
...
+ * Escapes ampersand, less-than, greater-than, single-quote, and double-quote
+ * characters with the corresponding Pango markup strings if markup is enabled.
+ *
+ */
+void maybe_escape_markup(char *text, char **buffer) {
+    if (markup_format == M_NONE) {
+        *buffer += sprintf(*buffer, "%s", text);
+        return;
+    }
+    else if (markup_format == M_PANGO) {
```
stapelberg: You can delete this conditional. If markup_format != M_NONE, it must be pango. This saves you one level of indention.

src/output.c
```diff
@@ -73,3 +73,39 @@ void print_separator(const char *separator) {
 void reset_cursor(void) {
     printf("\033[?25h");
 }
+
+/*
+ * Escapes ampersand, less-than, greater-than, single-quote, and double-quote
+ * characters with the corresponding Pango markup strings if markup is enabled.
```
stapelberg: Can you add a reference here to either pango’s definition of characters that need to be escaped, or, if that doesn’t exist, their official function, so that one can verify the two are in sync?

stapelberg:
> Only printing the wireless SSID has been updated. Is there any other generated content that might need escaping (i.e. containing &, <, >, ', ")?

As mentioned in #31, it might be prudent to do the same for the time module. Aside from that, I think we should also automatically escape all input provided by the user, e.g. format strings.

For that, I think it might be better to change maybe_escape_markup() to return a string instead of directly appending to a buffer.

ixjlyons:
> Aside from that, I think we should also automatically escape all input provided by the user, e.g. format strings.

I'm not sure about that. There will be special characters in the format string, e.g.:

```
format = "<span foreground='#eeeeee'>disk:</span> %avail"
```

Unless I've misunderstood something, I think users should be required to escape the special characters themselves (it's a minor inconvenience, at most). You could parse the markup and only escape things outside the `<span>` environment (or `<i>`, `<b>`, etc.), but that kind of complexity should preclude implementation of this feature in i3status, right?

stapelberg:
Fair enough, let’s skip escaping user input then.

src/output.c
ixjlyons: I'm having a little trouble finding something for this. The docs don't really say anything about escaping these characters. The link to the glib source you posted in #31 contains the version number in the url and it is broken.

Perhaps we need to go up a level to the XML specification -- would something like [this](http://www.w3.org/TR/xml/#syntax) be suitable?

stapelberg: I’d prefer to not refer to the XML specification, unless that’s how glib defines the escaping it performs (and I don’t think that’s the case).

https://git.gnome.org/browse/glib/tree/glib/gmarkup.c?id=03db1f455b4265654e237d2ad55464b4113cba8a#n2142 should be a more durable link to the source whose functionality we’re re-implementing.

src/output.c
```diff
@@ -73,3 +73,39 @@ void print_separator(const char *separator) {
...
+                    *buffer += sprintf(*buffer, "%s", "&gt;");
+                    break;
+                case '\'':
+                    *buffer += sprintf(*buffer, "%s", "&apos;");
+                    break;
+                case '"':
+                    *buffer += sprintf(*buffer, "%s", "&quot;");
+                    break;
+                default:
+                    *(*buffer)++ = *text;
```
stapelberg: What’s with this part of the code: https://git.gnome.org/browse/glib/tree/glib/gmarkup.c?id=03db1f455b4265654e237d2ad55464b4113cba8a#n2181 — do we not need it?

src/output.c
ixjlyons: I think it *is*, but I find it difficult to determine for sure. See the docs for [`g_markup_escape_text()`](https://developer.gnome.org/glib/unstable/glib-Simple-XML-Subset-Parser.html#g-markup-escape-text), which uses the function linked above. XML is mentioned. The link to the source for that function is much more clear, though.

src/output.c
ixjlyons: It seems to be replacing control characters with the XML code except tabstop, newline, and carriage return (see [`g_markup_escape_text()`](https://developer.gnome.org/glib/unstable/glib-Simple-XML-Subset-Parser.html#g-markup-escape-text)). I'm not sure if that is necessary for i3status, to be honest. It's simple to implement, if so.

stapelberg: I’m not 100% sure either — but if nothing else, it’s easier to compare that the functions do the same thing if we go the same route. Can you add it please?

ixjlyons:
> As mentioned in #31, it might be prudent to do the same for the time module.

The format string for the time section may contain markup (e.g. `format = "<b>time:</b> %H:%M"`), so this goes into and comes back out of `strftime()`. There is unfortunately nowhere in between to escape special characters in the *content* without also escaping the `<` and `>` characters in the markup tags.

The three options I can think of are:

1. Implement the `strftime()` character sequences so individual snippets of content can be escaped (as `%essid` in the wireless info module). This is obviously not the right way to go.

2. Escape everything except `<span>`, `<i>`, etc. This goes back to my comments above on automatically escaping all input, i.e. it seems too complex to warrant markup support in i3status.

3. Split up the format string specification in the config so there is a "pure" format string for passing to `strftime()` and a higher-level format string for markup, somewhat like the battery module (e.g. `time_str = "%H:%M"` / `format = "<b>time:</b> %time"`). I don't think this is a particularly good option, but it probably has a less complicated implementation than the other two.

Is there some other solution I'm not thinking of?

stapelberg:
I think option 3 is the best one — it’s most in line with other modules and requires the least implementation work. Please pay attention to making this backwards-compatible, i.e. don’t rename configuration options, but just add new ones.

ixjlyons:
I tried out option 3 for tztime, keeping backward compatibility in mind. Here are the cases and how they're treated:

* No markup: nothing is different. The string specified in `format` is passed to `maybe_escape_markup` but is unchanged since the global `markup` option is not set.

* Markup enabled, `format` option contains no markup (e.g. `format = "%H:%M %Z`): Works as above, but the output will be escaped.

* Markup enabled, markup desired within the format string: User will need to use the separate `format_time` option, e.g.:

```
format = "%time"
format_time = "%H:%M %Z"
```

ixjlyons:
@stapelberg: Just curious if you've had a chance to look at the update.

Sorry about the merge conflict by the way. I tested out merging into master and it's a trivial fix.

Airblader:
@stapelberg Please fix the conflict yourself by rebasing master onto your branch.

src/print_time.c
```diff
@@ -33,17 +33,37 @@ void set_timezone(const char *tz) {
...
 
-    /* Convert time and format output. */
     set_timezone(tz);
     localtime_r(&t, &tm);
-    outwalk += strftime(outwalk, 4095, format, &tm);
+
+    if (format_time == NULL) {
+        strftime(timebuf, sizeof(timebuf), format, &tm);
+        maybe_escape_markup(timebuf, &outwalk);
+    }
```
stapelberg: nit: please use `} else {` for consistency with the rest of the code base

stapelberg:
Looks good overall. Can you please rebase?

ixjlyons:
Don't know if you're notified when the PR gets updated with a push -- rebase complete, anyway.

stapelberg:
Thanks for your patience with this PR!

PR #73, opened 2015-10-10
Add ratio_watch to support displaying display brightness.

orbekk:
This is implemented by watching files in /sys/class/backlight. Tested on a Lenovo X1 with the intel_backlight module and linux 4.2.2.

### DIFF
--- a/i3status.c
+++ b/i3status.c
@@ -420,6 +420,18 @@ int main(int argc, char *argv[]) {
         CFG_CUSTOM_MIN_WIDTH_OPT,
         CFG_END()};
 
+    cfg_opt_t ratio_watch_opts[] = {
+        CFG_STR("format_down", "", CFGF_NONE),
+        CFG_STR("format_high", "☼: %percentage", CFGF_NONE),
+        CFG_STR("format_low", "☾: %percentage", CFGF_NONE),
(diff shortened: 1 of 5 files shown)


stapelberg:
Thanks for your pull request, but I’ll have to decline merging it for two reasons:

1. I don’t think this is a good fit for i3status as per its philosophy: display brightness is nothing that you would routinely query (compared to how much space you have available in your file system, or how much power your battery can still provide), largely because you can just perceive/see it. If your display is not bright enough, make it brighter. If it strikes you as to bright, make it less bright. There’s no point in knowing how bright it is.

2. Point ① is enough to make this decision. If it wasn’t, I’d prefer this functionality to be implemented as a specific backlight_brightness module. I can’t easily think of any other ratio that’d be important, and having such generic modules invites feature/scope creep, i.e. at some point, someone will think it’s a good idea to also multiply numbers with this module (because hey, it’s a two-line code change) and the module will get more confusing over time.

I suggest that, if you really care about this use-case, you either just patch i3status (easiest short-term solution, as you’ve already written the code) or implement this as a wrapper around i3status (most maintainable solution), see http://i3wm.org/i3status/manpage.html#_external_scripts_programs_with_i3status

PR #77, opened 2015-10-12
Correctly calculate Wireless link quality

StefaBa:
Wireless quality was stuck at 100% due to
a signed with unsigned integer division.

### DIFF
--- a/src/print_wireless_info.c
+++ b/src/print_wireless_info.c
@@ -104,7 +104,7 @@ static void print_bitrate(char *buffer, int buflen, int bitrate) {
 }
 
 // Based on NetworkManager/src/platform/wifi/wifi-utils-nl80211.c
-static uint32_t nl80211_xbm_to_percent(int32_t xbm, uint32_t divisor) {
+static uint32_t nl80211_xbm_to_percent(int32_t xbm, int32_t divisor) {
 #define NOISE_FLOOR_DBM -90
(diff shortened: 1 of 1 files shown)


stapelberg:
Can you explain why NetworkManager (which uses the same function signature) is not broken? See https://sources.debian.net/src/network-manager/1.0.6-1/src/platform/wifi/wifi-utils-nl80211.c/#L270

StefaBa:
honestly: not really. i'm currently looking at the CLAMP macro as this seems to be the only thing which differs from the i3status implementation.

StefaBa:
Ok, i don't know why NM works, because it should not.
glib type values have the same type-cast conventions as int32_t and uint32_t, there's no compiler flag which might reverse type-cast conventions (would be very stupid anyways). And CLAMP is just a blown up version of, well, a clamp function...

besides the calculation (if done correctly) is a bit weird anywas:
why return 100 - **70** * \<number between 0 and 1\>
and not 100 - 100 \<number between 0 and 1\>
(ok the then generated percent values would be lower than expected, but doesn't
that imply that the defined min/max dbm values are wrong...)

stapelberg:
I recall the driver reporting values between 0 and 70.

I’m a bit unsatisfied with not having an explanation as to why the two pieces of code aren’t equally broken/working, but that shouldn’t block this pull request.

PR #81, opened 2015-11-12
add local (mean) sidereal time

privong:
Based off the print_time() functionality, I added code to compute the [local mean sidereal time](https://en.wikipedia.org/wiki/Sidereal_time) and display it within i3status. Let me know if you have any feedback on this addition. If this is something you're interested in having as part of i3, I'm happy to iterate on it.

### DIFF
--- a/src/print_lst.c
+++ b/src/print_lst.c
@@ -22,7 +22,7 @@ void print_lst(yajl_gen json_gen, char *buffer, const char *title, const char *l
     now = time(NULL);
     d = (now / 86400.0) + 2440587.5 - 2451545.0; /* days since 2000 Jan 1 1200h UTC */
     t = d / 36525;                               /* centuries since 2000 Jan 1 1200h UTC */
-    GMST_s = 24110.54841 + 8640184.812866 * t + 0.093104 * t * t - 0.0000062 * t * t * t;
+    GMST_s = 24110.54841 + 8640184.812866 * t + 0.093104 * pow(t, 2) - 0.0000062 * pow(t, 3);
     GMST_s += now;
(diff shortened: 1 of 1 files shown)


privong:
I've created a slight variation on this, which puts the computed LST into a tm struct, so that the LST output can be customized using strftime(). It's a bit more flexible, but some of the time formatting options make less sense for LST than they do for normal time. e.g., there's no AM/PM. But, I can envision someone wanting only H:M of LST and not H:M:S, for example.

[This is the branch with the formatting option](https://github.com/privong/i3status/tree/lst_fmt). I can swap the pull request to that branch if you prefer the formatting option.

stapelberg:
Thanks for the pull request.

I’m happy to accept this into the contrib/ directory, but I think it’s too specific of a use-case for it to make sense in i3status itself.

Could you change the implementation to be a wrapper suitable for inclusion in contrib/ please?

privong:
@stapelberg More specific than the discordian date? ;)
Sure thing, it is a bit specialized. It might take me a bit before I get around to converting it to a contrib/ wrapper, though – I already have it integrated locally in i3status and my observing run is almost over :)

stapelberg:
Having discordian date in core i3status was a mistake, but removing it would break user’s configs, which we don’t do :). It definitely should have been added to contrib/ as well.

I’m not in a hurry, so take your time with the pull request.

Thanks!

Airblader:
I do think one year constitutes plenty of time, so closing this since nothing has happened.

privong:
@Airblader yeah. It's on my long "to do" list, but I'd been using my modded copy locally, so I didn't have too much incentive to turn it into a wrapper. I'll open a new request if I ever get around to rewriting it. :)

PR #86, opened 2015-12-23
update README.md with libnl-genl-3-dev

remijouannet:
hi, simple PR to update the README.md,
I had to install libnl on my desktop to compile i3status and it was missing from the readme


remijouannet:
(well it was not missing from the bullet list but it was missing from the command line)

PR #89, opened 2015-12-29
add support for reading text files

krgn:
Hello,

I would like to propose adding support for reading and displaying file contents in i3status. I can be used to read in cached data from disk, like mail counts or statistics in plain text.

Please let me know if there are any improvements I should make to the code.


stapelberg:
Duplicate of https://github.com/i3/i3status/pull/19, please see the rationale there as to why we’re not going to merge such pull requests.

PR #94, opened 2016-01-11
Forgotten wakeup call from USR1 handler

### LINKED ISSUE #93: i3status 2.10 doesn't refresh when sent SIGUSR1
As the title says, the refresh doesn't work anymore for me. I'm using the Ubuntu wily and the stable repo mentioned on i3wm.org. But I can also reproduce the bug with a version I built from source.

How can I go about debugging this?

rpls:
Fixes #93 

I'm not a pthread expert. Is this the right way to do it? It fixed the issue for me.

### DIFF
--- a/i3status.c
+++ b/i3status.c
@@ -82,6 +82,7 @@ void fatalsig(int signum) {
  *
  */
 void sigusr1(int signum) {
+    pthread_cond_broadcast(&i3status_sleep_cond);
 }
 
 /*


Watcom:
I'm not a pthread expert either, but quoting from the manpage:
> "It is not safe to use the pthread_cond_signal() function in a signal handler that is invoked asynchronously. Even if it were safe, there would still be a race between the test of the Boolean pthread_cond_wait() that could not be efficiently eliminated.

>  Mutexes and condition variables are thus not suitable for releasing a waiting thread by signaling from code running in a signal handler."

That being said, I don't know of any alternative solution. Of course, this is only an issue when (and if) you actually use USR1.

rpls:
True. I did use the broadcast method though, don't know how much of a difference it makes. The restriction doesn't mention broadcast, but I guess the second paragraph kind of includes it too. Anyway, why was `nanosleep` removed? Doesn't this work with the PulseAudio patch?

Watcom:
The change from `nanosleep` was so PulseAudio events could wake up the main thread (i.e. on volume changes).

stapelberg:
Sounds to me like we should find a better solution, then. Maybe we can make use of signalfd in some way?

Watcom:
I think `signalfd` is Linux-specific, though.

Perhaps the simplest way is to go back to using `nanosleep()` and then use `pthread_kill()` to send USR1 to the main thread from `pulse.c`, dropping the condition variable altogether. I haven't tried this but in theory it should work.

I see [other solutions](http://osiris.978.org/~alex/safesignalhandling.html), but all of them seem to add more complexity.

stapelberg:
Going back to `nanosleep` as you suggested sounds good. Can you send a PR for that please?

PR #97, opened 2016-01-12
adding uninstall to makefile

remijouannet:
Just adding uninstall to the MakeFile


stapelberg:
We don’t want to support an uninstall target in i3 and related tools. The rationale is that people should be using their distribution packages, not installing/uninstalling from source.

PR #99, opened 2016-01-17
Add command for custom shell commands

dopefishh:
This adds the possibility to show the output of a given shell command in the status output without the need no wrap `i3status`.


Airblader:
Sorry, but we don't accept such pull requests for i3status. Please see #66 for the reasoning behind this.

PR #102, opened 2016-01-24
Colorized output for cpu temp and battery info

mfulz:
With this patch you're able to change the colors for cpu_temperature and battery with using color_good directives.

### DIFF
--- a/src/pulse.c
+++ b/src/pulse.c
@@ -13,7 +13,8 @@
 typedef struct indexed_volume_s {
     uint32_t idx;
     int volume;
-    TAILQ_ENTRY(indexed_volume_s) entries;
+    TAILQ_ENTRY(indexed_volume_s)
+    entries;
 } indexed_volume_t;
 
 static pa_threaded_mainloop *main_loop = NULL;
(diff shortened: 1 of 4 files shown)


stapelberg:
Please don’t change unrelated files, and please squash your fixup commits. Let me know when you’ve gotten the pull request to pass travis and i’ll take another look.

src/pulse.c
```diff
@@ -21,7 +22,8 @@ static pa_context *context = NULL;
 static pa_mainloop_api *api = NULL;
 static bool context_ready = false;
 static uint32_t default_sink_idx = DEFAULT_SINK_INDEX;
-TAILQ_HEAD(tailhead, indexed_volume_s) cached_volume =
+TAILQ_HEAD(tailhead, indexed_volume_s)
```
Airblader: @stapelberg If this is what a recent clang does with the current configuration, we should probably still change the configuration.

mfulz: I'll cleanup the code and hell fixing the formatting.
I just wanted top know, if the feature would be accepted before putting top mich work Ingo it.
I've used clang-format Version 3.7 for the formatting.

mfulz:
I switched to i3blocks -> no need for me.

PR #103, opened 2016-01-25
This removes the 'remaining' time when the battery is full, since the…

craftyguy:
… remaining time is always 00:00:00. When the battery is discharging, the remaining time will show up again if the user has set their format correctly.

### DIFF
--- a/src/print_battery_info.c
+++ b/src/print_battery_info.c
@@ -152,7 +152,7 @@ void print_battery_info(yajl_gen json_gen, char *buffer, int number, const char 
         (void)snprintf(percentagebuf, sizeof(percentagebuf), "%.02f%s", percentage_remaining, pct_mark);
     }
 
-    if (present_rate > 0) {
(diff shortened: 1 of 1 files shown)


craftyguy:
The Travis Cl build environment is all screwed up. I tried using clang-format to correct some formatting issues that I DIDN'T create but the build is still screwed up. I don't think my one line change is at fault since the build env can't find basic includes..

stapelberg:
Restarting the travis build made it work again.

Can you please revert the clang-format change commit? It seems like you’re using a different version of clang-format. We’re using clang-format-3.5.

craftyguy:
Sorry for the thrash.. I reverted the clang-format commit

stapelberg:
Thanks. Can you also squash the commits please? After merging, our history doesn’t need to capture the two commits that cancel each other out, we’re only interested in the actual change you’re making.

craftyguy:
Ok, I think I got it now! Thanks for the patience!

PR #107, opened 2016-02-01
Implement CPU frequency display for Linux and OpenBSD.

weezel:
This patch adds support for showing (an average) frequency of all CPUs.

Unfortunately **this doesn't work yet** and I fail to spot what I am missing, since the frequency column stays blank.

Tried to test this way:
i3config.conf:
`+  order += "cpu_freq"`

$ make clean all
$ ./i3status -c i3config.conf

but the frequency column stays blank. Tips welcome.

Thanks

### DIFF
--- a/i3status.c
+++ b/i3status.c
@@ -390,6 +390,12 @@ int main(int argc, char *argv[]) {
         CFG_CUSTOM_MIN_WIDTH_OPT,
         CFG_END()};
 
+    cfg_opt_t cpufreq_opts[] = {
+        CFG_STR("format", "%MHz", CFGF_NONE),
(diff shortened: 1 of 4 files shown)


Airblader:
Sorry, but we don't accept such PRs into i3status. Sorry about the wasted work, but we generally recommend to ask about features before implementing them to get feedback on acceptance and design. This is especially true for i3status where we are rather restrictive.

See also the man page of i3status for an explanation (or b35753cd)

weezel:
Sure, no problem and thanks for the pointer. I was mainly doing that for my own amusement. Any tips what I did wrong since I'd like to use that in a local repository?

PR #113, opened 2016-03-10
Correct i3status.man load path for /etc/i3status/config

grahamc:


grahamc:
Closing, because I was confused. Sorry for the noise.

Issue #25, opened 2015-04-12
Wireless %quality has leading zero when quality ~=100


dougmercer:
This is a very minor nitpick, but I'm not quite sure how to trim the leading zero from the %quality string when quality is ~=100. The reason I'm submitting this an issue and not simply assuming that it's intended behavior is because volume and battery life seem to correctly trim the zero. 

Here is an illustration of what I'm referring to http://i.imgur.com/9qhKV1P.jpg

As an aside, thank you for i3 and all of its related software. I'm just beginning to use it and am already hooked.

stapelberg:
We’ve done this since commit 71f71660322b12b8602f5979222d280b7db333bd (June 2010) to avoid the text jumping around when your wireless quality oscillates between 100% and <100%.

You could send a pull request to introduce a new format string (format_quality) which makes this configurable.

dougmercer:
Ah, that makes sense. I'll try to to make this a configurable setting once this semester is finished.

Thanks for the quick response!

neiljp:
I found this problem myself, so had a play with fixing it myself.
Would you consider #115 as a fix for this?

PR #115, opened 2016-03-16
Add & document quality_leading_character option for wireless

neiljp:
I added this function to my local version, and it appears to be a fix for #25.

I don't currently have the capability to test the manpage (or changes), but the markdown is fairly straightforward so can't see it should be a problem.

The only significant unknown is whether to hoist the extra wireless code out of the info.quality_max test, but I'm not 100% sure when the other branch of the if is currently triggered, so have left it alone for now.

### DIFF
--- a/i3status.c
+++ b/i3status.c
@@ -328,6 +328,7 @@ int main(int argc, char *argv[]) {
     cfg_opt_t wireless_opts[] = {
         CFG_STR("format_up", "W: (%quality at %essid, %bitrate) %ip", CFGF_NONE),
         CFG_STR("format_down", "W: down", CFGF_NONE),
+        CFG_STR("quality_leading_character", NULL, CFGF_NONE),
         CFG_CUSTOM_ALIGN_OPT,
         CFG_CUSTOM_COLOR_OPTS,
         CFG_CUSTOM_MIN_WIDTH_OPT,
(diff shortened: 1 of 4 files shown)


stapelberg:
Instead of implementing padding with configurable characters, I’d prefer it if the format that’s passed to `sprintf` was made configurable (as indicated in #25). That way, users could easily get adjustment (`%-3d`) or no padding at all (`%d`).

What do you think?

neiljp:
I think I now follow your original comment in #25. I'm not tied to the idea of 'other characters', so the format string approach could functionally work in terms of selecting between padding with spaces or zeroes or...unpadded.

However, I don't think that exposing a fundamental format string in the config file (and expecting users to know/learn them) is the best way to go. It would make it "very" configurable, but I honestly don't see what it would gain users: what does a width of less than 3 mean for a %? more than 3? (equivalent to other text in the main format string!) what if they use %f? what strings are supported?

Thinking about it more generally, though not having looked at all the potentially impacted outputs, it could be useful for each % output (wifi quality, battery (?), volume, ...) accept a percent_padded (true/false) and optional related percent_padding_character (default to space, or zero?)

stapelberg:
Here’s the test program I used to answer your questions below:
```c
#include <stdio.h>

int main() {
    int val = 60;
    printf("%%d: %d\n", val);
    printf("%%01d: %01d\n", val);
    printf("%%02d: %02d\n", val);
    printf("%%03d: %03d\n", val);
    printf("%%04d: %04d\n", val);
    printf("%% 4d: % 4d\n", val);
    printf("%%-4d: %-4d\n", val);
    printf("%%f: %f\n", val);
}
```

> what does a width of less than 3 mean for a %?

`%01d` results in the same output as `%03d`.

> more than 3? 

Higher numbers results in more padding: `%02d` results in `60`, `%03d` results in `060`, `%04d` results in `0060` and so on.

> what if they use %f?

On my machine, this results in `0.000000`, but I guess this depends on the precise value and possibly the C implementation. In any case, it’s not supported :).

> what strings are supported?

`%d` only.

We already have configuration options that are directly passed to libc functions, see e.g. the `time` module’s `format`, which is being passed to `strftime()`.

I think we should use format strings here as well, if only for consistency. Personally, I prefer directly modifying the format string over a number of additional options which achieve the same thing (known concept vs. i3status-specific configuration).

Airblader:
No response in half a year, closing this.

PR #117, opened 2016-03-21
Added status_unk to battery_info

egroat:
Some batteries report "Unknown" for battery status where they have some form of alternate mode.  Mine for example switches to AC supply when the battery is at 80% (BIOS option for extended battery life) and a colleagues allows the use of an external battery.

As such the existing i3status reports discharging when it is not the case.  As there is unlikely to be more than one "unknown" mode per battery this modification adds a battery option of "status_unk"  with default %status value of "UNK" allowing users to identify and configure for their battery.

This will have different uses for different people - I  set it to show I am plugged in and not charging,  my colleague will show external (but not wall) power,  and many others would derive no benefit whatsoever.

Note:

  This has only been implemented for the Linux ifdef.  If this feature is not desired in the main repo there is no point adding support for systems I do not use;  if this is something that is desired I am able to source readouts / work on the code for the other systems (though do not have copies running).

### DIFF
--- a/i3status.c
+++ b/i3status.c
@@ -360,6 +360,7 @@ int main(int argc, char *argv[]) {
         CFG_STR("format_down", "No battery", CFGF_NONE),
         CFG_STR("status_chr", "CHR", CFGF_NONE),
         CFG_STR("status_bat", "BAT", CFGF_NONE),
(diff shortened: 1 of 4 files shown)


memeplex:
I think the change I proposed in https://github.com/i3/i3status/issues/119 will make a lot of sense for this particular status.

PR #122, opened 2016-04-13
Add CPU usage color thresholds

mark7:
CPU usage had previously not supported the color option.  Add support
for a "degraded" state above which the degraded color is used, and a
higher "bad" state above which the "bad" color is used.  One possible
use for these might be indicating whether one or all cores are
saturated.

Unlike the color settings for other, these are set high enough to be
disabled by default.  This is done because i3status determines CPU
usage over only the last display interval, which means that, a user
with a low refresh rate might see frequent, potentially-annoying color
changes.

### DIFF
--- a/man/i3status.man
+++ b/man/i3status.man
@@ -381,11 +381,24 @@ specified thermal zone is getting too hot. Defaults to 75 degrees C.
 
 === CPU Usage
 
-Gets the percentual CPU usage from +/proc/stat+ (Linux) or +sysctl(3)+ (FreeBSD/OpenBSD).
+Gets the percentual CPU usage from +/proc/stat+ (Linux) or +sysctl(3)+
+(FreeBSD/OpenBSD).
+
+It is possible to define a max_threshold that will color the load
+value red in case the CPU average over the last interval is getting
+higher than the configured threshold. Defaults to 200 (i.e. off).
+
+It is possible to define a degraded_threshold that will color the load
+value yellow in case the CPU average over the last interval is getting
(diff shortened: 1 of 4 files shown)


man/i3status.man
```diff
@@ -381,12 +381,25 @@ specified thermal zone is getting too hot. Defaults to 75 degrees C.
...
+Gets the percentual CPU usage from +/proc/stat+ (Linux) or +sysctl(3)+
+(FreeBSD/OpenBSD).
+
+It is possible to define a max_threshold that will color the load
+value red in case the CPU average over the last interval is getting
+higher than the configured threshold. Defaults to 200 (i.e. off).
+
+It is possible to define a degraded_threshold that will color the load
+value yellow in case the CPU average over the last interval is getting
+higher than the configured threshold.  Defaults to 150 (i.e. off).
```
stapelberg: The code actually says 200, not 150. Can you fix the docs please?

man/i3status.man
```diff
@@ -381,12 +381,25 @@ specified thermal zone is getting too hot. Defaults to 75 degrees C.
...
 
 *Example order*: +cpu_usage+
 
 *Example format*: +%usage+
 
+*Example max_threshold*: +"200,0"+
+
+*Example degraded_threshold*: +"150,0"+
```
stapelberg: Both examples use a local-specific separator. Can you just use integer values in the examples please?

Also, specifying 200 and 150 won’t do anything AIUI. Can you update the examples to actually do something useful please?

stapelberg:
Closing in favor of #139

PR #125, opened 2016-05-05
reconnect to PulseAudio if connection is lost (fixes #124)

### LINKED ISSUE #124: Volume doesn't get updated if pulseaudio is restarted
After running `pulseaudio -k && pulseaudio --start`, i3status stops updating the display for the volume.

Watcom:
Just tested it, seems to fix the problem.


stapelberg:
@nicooo Does this patch fix the issue for you?

nicooo:
Yes

PR #127, opened 2016-05-13
Add ifdef to compile wireless_info on Mac

afh:


afh:
Thanks for merging so quickly! 👍

PR #134, opened 2016-06-30
Manpage: Subsection tzdata: Example config now matches to context

Bandie:
When I tried to configure a tzdata section to myself I noticed that the example configuration wasn't clear enough in the given context. 

I suggest this as a little improving.


stapelberg:
Thanks!

Issue #138, opened 2016-07-23
feature: aggregate representation for multiple batteries


dustymabe:
Some newer laptops have multiple batteries. rather than modifying the config to show both batteries it would be cool to have support (could be default behavior or non-default configurable) for showing both batteries as an aggregate. This would treat them as 1 and show the capacity of them together and the actual remaining time til power outage. 

Here is an example where someone has a manual workaround for this:
https://josiah.wordpress.com/2015/01/29/making-i3status-work-with-multiple-batteries/

Maybe we could get that type of functionality in i3 if we think it's worthy. 

Thanks for the great work!

stapelberg:
I think it’s unfortunately not so simple to merge the status of multiple batteries.

As an example, i3status can only predict the remaining battery time by looking at the remaining capacity and the discharge rate. However, discharge rate is only reported when a battery is actually discharging, and only one battery is discharging at a time. Hence, the remaining time would never be accurate. One could make the assumption that both batteries should discharge at roughly the same rate, but I’m hesitant to hard-code that assumption into i3status, as I’m not convinced that it typically holds.

dustymabe:
Thanks @stapelberg. That makes sense. What do you think of the way the guy is doing it in https://josiah.wordpress.com/2015/01/29/making-i3status-work-with-multiple-batteries/ by combining the uevent information from both batteries. Does his method suffer from that bad assumption?

stapelberg:
Given that we’re talking about a rather uncommon case (the vast majority of notebooks have precisely 1 battery), I think for now I prefer to not add that complexity (both technical, and, more importantly, mental) to i3status.

Airblader:
I +1 the request, though I agree with the issue. I also don't use i3status myself, but I have the very situation that my machine has two batteries.

PR #140, opened 2016-08-01
Implement aggregates in print_battery_info

tommie:
This allows Linux and NetBSD code to use the magic "battery -1" to show an aggregate of all batteries. It is useful e.g. on the Thinkpad T460s, which has two batteries.

A substantial amount of refactoring to make the code manageable and remove duplication between OSes. Hopefully split into sensible commits. I can't guarantee it preserves existing behavior, but hopefully it improves it rather than makes it worse.

Tested in Gentoo Linux on a T460s and FreeBSD, OpenBSD, NetBSD in VirtualBox. VirtualBox doesn't emulate ACPI fully, so I can't say e.g. the time estimates work. The code compiles and responds to state/percentage changes at least. The NetBSD code didn't even compile cleanly, so that's an improvement.

Changed files have been run through clang-format-3.7 -style=file.

### DIFF
--- a/i3status.conf
+++ b/i3status.conf
@@ -15,7 +15,7 @@ order += "ipv6"
 order += "disk /"
 order += "wireless _first_"
 order += "ethernet _first_"
-order += "battery 0"
+order += "battery all"
 order += "load"
 order += "tztime local"
 
@@ -30,7 +30,7 @@ ethernet _first_ {
         format_down = "E: down"
 }
 
-battery 0 {
+battery all {
         format = "%status %percentage %remaining"
 }
 
--- a/src/print_battery_info.c
+++ b/src/print_battery_info.c
@@ -9,6 +9,12 @@
 
 #include "i3status.h"
 
+#if defined(LINUX)
+#include <errno.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#endif
+
 #if defined(__FreeBSD__) || defined(__FreeBSD_kernel__) || defined(__DragonFly__)
 #include <sys/types.h>
 #include <sys/sysctl.h>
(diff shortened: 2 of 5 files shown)


Airblader:
> This allows Linux and NetBSD code to use the magic "battery -1" to show an aggregate of all batteries.

How does it deal with the problem mentioned in #138? Have you seen that issue? We decided just a week ago to not implement aggregation in i3status, so unless you bring something new to the table, this PR will most likely be rejected.

> I can't guarantee it preserves existing behavior.

That's really not a good basis for getting a change merged. Perhaps this means the change is too big to be merged without a test suite at the moment.

> src/print_battery_info.c has been run through clang-format-3.7 -style=file.

We use clang-format3.5 and in my experience there really are differences in formatting between the versions (despite using a style file).

i3status.conf
```diff
@@ -17,7 +17,7 @@ order += "run_watch DHCP"
 order += "run_watch VPN"
 order += "wireless _first_"
 order += "ethernet _first_"
-order += "battery 0"
+order += "battery -1"
```
Airblader: I'm not a fan of such magic numbers. This should, if at all, be done differently, IMHO.

tommie:
> How does it deal with the problem mentioned in #138? Have you seen that issue? We decided just a week ago to not implement aggregation in i3status, so unless you bring something new to the table, this PR will most likely be rejected.

If you are referring to

> One could make the assumption that both batteries should discharge at roughly the same rate, but I’m hesitant to hard-code that assumption into i3status, as I’m not convinced that it typically holds.

What's described there is a system that would change power usage when switching batteries. To me, that seems like an uncommon case. Most "ultra portables" only have two batteries so they can be fitted in multiple free spaces. They are still used to power the same load no matter which battery is active. So, summing the capacities before doing a ratio should be enough for the common case. Using the aggregates is optional, so if you have load that changes based on what battery is used, then you can fall back to individual gauges. Making the assumption that the machine will run until all batteries are depleted doesn't sound like a stretch.

Note that FreeBSD and OpenBSD already ignores the "number" and presumably returns an aggregate (or just returns an arbitrary battery stat).

>> I can't guarantee it preserves existing behavior.
> That's really not a good basis for getting a change merged. Perhaps this means the change is too big to be merged without a test suite at the moment.

Agreed.

>> src/print_battery_info.c has been run through clang-format-3.7 -style=file.
> We use clang-format3.5 and in my experience there really are differences in formatting between the versions (despite using a style file).

Ok.

stapelberg:
> What's described there is a system that would change power usage when switching batteries.

IIUC, that statement is making the assumption that the reported discharge rate _is_ the actual discharge rate, and not skewed on a per-battery basis, which might be the case, especially when different vendors are involved.

Currently, i3status uses battery-reported values in relation to each other, not across batteries.

stapelberg:
Please move the unrelated changes (making the NetBSD code compile cleanly etc.) into separate pull requests, so that we can make a distinction between fixes, no-op changes and actual changes.

tommie:
> IIUC, that statement is making the assumption that the reported discharge rate is the actual discharge rate, and not skewed on a per-battery basis, which might be the case, especially when different vendors are involved.

Ah, you're saying the precision will be worse because of uncalibrated measurements. Yepp, I can buy that. In my case, the system depletes one battery at a time, so less of an issue since I usually have the case that one battery is either complete full or completely empty with a power usage of zero, out of the two.

> Please move the unrelated changes (making the NetBSD code compile cleanly etc.) into separate pull requests, so that we can make a distinction between fixes, no-op changes and actual changes.

Will do.

i3status.conf
tommie: I'll use the (magic, but descriptive) string "all" instead.

tommie:
All prerequisite refactoring commits merged. Thanks!

I need to improve the accumulation code before moving forward. It doesn't deal well with the theoretical case of one battery charging while another is discharging.

Airblader:
I don't actually use i3status, but I'll probably end up using i3status solely for battery information in my i3blocks setup once this is merged. Thanks again!

tommie:
I think this version should be a good start. The merging of statuses is a bit complex since the aggregate of present_rate may contradict the statuses. In that case it is assumed present_rate has better information. For single-battery systems, it will just be merging UNKNOWN with something else, so is hopefully trivially correct.

src/print_battery_info.c
```diff
@@ -28,17 +34,81 @@
...
+        acc->full_last = 0;
+        acc->remaining = 0;
+        acc->present_rate = 0;
+    }
+
+    acc->full_design += batt_info->full_design;
+    acc->full_last += batt_info->full_last;
+    acc->remaining += batt_info->remaining;
+
+    /* make present_negative negative for discharging and positive for charging */
```
stapelberg: s/present_negative/present_rate/

src/print_battery_info.c
```diff
@@ -28,17 +34,81 @@
...
+    int present_rate = (acc->status == CS_DISCHARGING ? -1 : 1) * acc->present_rate;
+    present_rate += (batt_info->status == CS_DISCHARGING ? -1 : 1) * batt_info->present_rate;
+
+    /* merge status */
```
stapelberg: Can you clarify in the comment how the code below works? I.e., why does it look at the previous `acc->present_rate`, not the just calculated `present_rate`?

src/print_battery_info.c
tommie: Done.

src/print_battery_info.c
tommie: Let's call it a bug. Fixed.

stapelberg:
@Airblader I lack the hardware to test this. Could you take a look and let me know if this PR is good to go?

@tommie Could you rebase please?

tommie:
> @tommie Could you rebase please?

Done.

Airblader:
I'm using the following config file:

````
# i3status configuration file.
# see "man i3status" for documentation.

# It is important that this file is edited as UTF-8.
# The following line should contain a sharp s:
# ß
# If the above line is not correctly displayed, fix your editor first!

general {
        colors = true
        interval = 1
}

order += "battery 0"
order += "battery 1"
order += "battery all"

battery 0 {
        format = "%status %percentage %remaining"
        path = "/sys/class/power_supply/BAT0/uevent"
        last_full_capacity = true
}

battery 1 {
        format = "%status %percentage %remaining"
        path = "/sys/class/power_supply/BAT1/uevent"
        last_full_capacity = true
}

battery all {
        format = "%status %percentage %remaining"
        path = "/sys/class/power_supply/BAT%d/uevent"
        last_full_capacity = true
}
```

And ran i3status with the charger disconnected. After a while the display reads:

````
UNK 95.56% | BAT 73.50% 00:43:09 | UNK 95.56%
```

This doesn't sound right to me. The combined battery is the same as the first one even though the second one is only 3/4 full. Am I doing something wrong?

tommie:
It also shouldn't say unknown since at least one battery is not unknown. If I run the same I see

    UNK 94.42% | UNK 100.00% | UNK 97.57%

which seems right.

Are you sure you're running with the latest version? I'd think that the previous version that didn't parse "all" would return 0 if it was passed non-digits.

Airblader:
> Are you sure you're running with the latest version?

*head → wall* I think you're right. I was running master, but of course this isn't merged yet. My bad :-( If only `all` would've errored out… :-) I'll try it again tomorrow when I have access to that laptop again.

Airblader:
Tried it again, this time with the correct branch. :-) Everything seems to work fine for me now as well!

PR #142, opened 2016-08-02
Remove statusbuf, percentagebuf and remainingbuf from print_battery_info

tommie:
In preparation for #140, this uses raw values to transfer data from retrieval to formatting.

The only intended functional change is that all OSes now show the "(until full)" when charging. Previously only NetBSD did so.

I suggest reviewing commit by commit, but lumped them together in a PR since they are small. There are two more buffers to remove after this, but they have more *BSD functional changes.

### DIFF
--- a/src/print_battery_info.c
+++ b/src/print_battery_info.c
@@ -38,7 +38,6 @@ void print_battery_info(yajl_gen json_gen, char *buffer, int number, const char 
     time_t empty_time;
     struct tm *empty_tm;
     char buf[1024];
-    char percentagebuf[16];
     char remainingbuf[256];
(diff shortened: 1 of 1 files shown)


stapelberg:
> The only intended functional change is that all OSes now show the "(until full)" when charging. Previously only NetBSD did so.

I think the fact that NetBSD has that suffix was an oversight when merging NetBSD support. Could you please change it so that no operating system has that suffix? The vast majority of our users use Linux, and I’d like to avoid unexpected behavior changes like this one.

tommie:
Removed the "until full" suffix.

PR #146, opened 2016-08-06
Deleting VERSION and extracting it from I3STATUS_VERSION instead

eplanet:
I propose this little change for easier version management with only one file containing version number. VERSION is redundant since I3STATUS_VERSION contains the same amount of information.


stapelberg:
Can you make the same change for https://github.com/i3/i3 please? I’d like to have the same mechanism for all repositories.

eplanet:
On my way :-)

PR #148, opened 2016-08-08
Not mounted disk without format_not_mounted printing dash. Fixes #147

### LINKED ISSUE #147: Unmounted disk prints wrong information
While we can easily specify an error message (or whatever value) through `format_not_mounted` option, when we don't, random wrong values are printed for unmounted disk:
```yaml
disk "/unmounted/disk" {
    format = "unmounted/disk: %used (%avail)/ %total"
}
```
Result in:
```bash
$ ./i3status
unmounted/disk: 5.9 MiB (128.0 TiB) / 5.9 MiB
```
I think we should print empty values (or maybe "-"), when `format_not_mount` is not specified and the disk is not mounted, to end up with something like:
```bash
unmounted/disk: - MiB (- TiB)/ - MiB
```
I'd be happy to propose a PR to implement that if you agree.

eplanet:
I am not immensely happy with the verbosity of the if / else _en masse_, but here's a proposition that work fine with the following configuration file:
```yaml
order += "disk /"
order += "disk /notmounted/disk"
order += "disk /notmounted/disk/format"

disk "/" {
        format = "/ '%free' '%used' '%total' '%avail' '%percentage_free' '%percentage_used_of_avail' '%percentage_used' '%percentage_avail'"
}

disk "/notmounted/disk" {
        format = "/notmounted/disk '%free' '%used' '%total' '%avail' '%percentage_free' '%percentage_used_of_avail' '%percentage_used' '%percentage_avail'"
}

disk "/notmounted/disk/format" {
        format = "'%free' '%used' '%total' '%avail' '%percentage_free' '%percentage_used_of_avail' '%percentage_used' '%percentage_avail'"
    format_not_mounted = "/notmounted/disk/format is not mounted…"
}
```
Here's a sample output:
```bash
$ ./i3status -c notmounteddisk.conf 
/ '19.0 GiB' '4.4 GiB' '23.4 GiB' '17.8 GiB' '81.1%' '24.0%' '18.9%' '76.0%' | /notmounted/disk '–' '–' '–' '–' '–' '–' '–' '–' | /notmounted/disk/format is not mounted…
```
I also took the liberty of adding a the function `print_percentage` to factor the constants used in the related sprintf. I can remove that if you prefer that the PR sticks to the point.

### DIFF
--- a/man/i3status.man
+++ b/man/i3status.man
@@ -249,9 +249,7 @@ is assumed to be "percentage_avail" and low_threshold to be set to 0, which
 implies no coloring at all.
 
 You can define a different format with the option "format_not_mounted"
-which is used if the path does not exist or is not a mount point. So you can just empty
-the output for the given path with adding +format_not_mounted=""+
-to the config section.
+which is used if the path does not exist or is not a mount point. Defaults to "".
 
 *Example order*: +disk /mnt/usbstick+
 
--- a/src/print_disk_info.c
+++ b/src/print_disk_info.c
@@ -130,13 +130,15 @@ void print_disk_info(yajl_gen json_gen, char *buffer, const char *path, const ch
 #else
     struct statvfs buf;
 
+    if (format_not_mounted == NULL) {
+        format_not_mounted = "";
+    }
+
     if (statvfs(path, &buf) == -1) {
         /* If statvfs errors, e.g., due to the path not existing,
          * we use the format for a not mounted device. */
-        if (format_not_mounted != NULL) {
-            format = format_not_mounted;
-        }
-    } else if (format_not_mounted != NULL) {
+        format = format_not_mounted;
+    } else {
         FILE *mntentfile = setmntent("/etc/mtab", "r");
         struct mntent *m;
         bool found = false;


Airblader:
I also don't like the if / else cascades too much (on a smaller note, please use braces even for one-line statements). Why not make printing the dash the responsibility of the individual `print_*` functions?

Perhaps an alternative would also be to just provide a default value for `format_not_mounted`. That'd be the easiest solution.

eplanet:
The default value for `format_not_mounted` is actually so obvious, it must be the best option here.
Do you have any idea for the value ? What do you think of "$diskname: not mounted" ?

Airblader:
In general I'd prefer a language agnostic default value, though I don't think it matters much here. But maybe `format_not_mounted` should just default to empty string? We could even get rid of the entire `format_not_mounted == NULL` handling (I think).

(But please wait for @stapelberg's input here)

stapelberg:
Looks good overall.

Can you please update the commit message to reflect what the code does? Something like “Print empty string for not mounted disks” with a more verbose explanation in the commit description.

eplanet:
I changed the commit message. I don't think it's necessary to add anything since the modification is quite small at the end. Is is ok for you ?

stapelberg:
Please move the “fixes #147” into the commit description to be consistent with our other commits (also, I’m not certain whether GitHub would parse it correctly when it’s at the end of the title).

I still think it makes sense to expand on the previous and new behavior in the commit message, since the title alone might be misleading otherwise.

eplanet:
I added some information to the commit message. And I actually realized it should be nice to update the man page as well since the default value is now "" for `format_not_mounted`.

PR #149, opened 2016-08-14
Fix some compilation warnings on OpenBSD and NetBSD

tommie:
This was a by-the-way on the path to #140.

These are the last commits with general improvements to the *BSD code. They should have no functional impact.


Airblader:
Just FYI, I think it's great that you're providing all these improvements. Thanks for the contributions!

PR #153, opened 2016-08-18
Parse uevent file to detect network device type

cakturk:
Currently i3status differentiates wireless and wired devices based
on the existence of wireless directory inside the device's sysfs
directory.  This approach seems to cause 3g modems to be incorrectly
identified as the first ethernet device.

Use DEVTYPE variable from uevent file to identify network interface
type.

### DIFF
--- a/src/first_network_device.c
+++ b/src/first_network_device.c
@@ -6,11 +6,58 @@
 
 #include "i3status.h"
 
+static bool sysfs_devtype(char *dest, size_t n, const char *ifnam) {
+    FILE *fp;
+    char buf[1024];
+
+    snprintf(buf, sizeof(buf), "/sys/class/net/%s/uevent", ifnam);
+    if ((fp = fopen(buf, "r")) == NULL)
+        return false;
+
+    dest[0] = '\0';
+
+    while (fgets(buf, sizeof(buf), fp)) {
+        size_t slen;
+        char *s;
+
+        slen = strlen(buf);
+        /* Line is too long to fit in the buffer */
+        if (buf[slen - 1] != '\n' && !feof(fp))
+            break;
+        if ((s = strchr(buf, '='))) {
+            if (strncmp(buf, "DEVTYPE", s - buf))
+                continue;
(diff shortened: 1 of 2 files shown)


src/first_network_device.c
```diff
@@ -6,11 +6,60 @@
 
 #include "i3status.h"
 
+#define streq(s1, s2) (strcmp(s1, s2) == 0)
```
stapelberg: Please remove this macro — elsewhere in the code we use the `strcmp() == 0` pattern, so let’s make the new code consistent.

src/first_network_device.c
```diff
@@ -6,11 +6,60 @@
...
+
+static bool sysfs_devtype(const char *ifnam, char *value, size_t n) {
```
stapelberg: nit: I think it’d be more idiomatic (see e.g. `snprintf`) to have the destination and size parameters come first, i.e.:

```c
static bool sysfs_devtype(char *dest, size_t n, const char *ifnam) {
```

stapelberg:
Looks good overall.

src/first_network_device.c
cakturk: Ok I'll remove the macro.

src/first_network_device.c
cakturk: Makes sense to me.

cakturk:
Hi @stapelberg, I've updated the PR as you suggested.

earboxer:
@cakturk after this PR, is there a way to see the status of a wireless modem? I've tried creating a `wireless wwan0` block, but that doesn't seem to work.

cakturk:
Hi @earboxer,

Can you paste the output of "cat /sys/class/net/\<device\>/uevent"? Replace \<device\> with the name of the corresponding wireless modem.

earboxer:
user$ cat /sys/class/net/wwan0/uevent
DEVTYPE=wwan
INTERFACE=wwan0
IFINDEX=4

PR #156, opened 2016-08-19
Add ability to find interfaces with partial name matches

howlett:
Some tethering devices change their interface name on reboot or
creation.  This allows finding the first device with a substring of the
name by checking for the special string _first_<DEV NAME>.  _first_
without a partial <DEV NAME> is unchanged.

Signed-off-by: Liam R. Howlett <howlett@gmail.com>

### DIFF
--- a/src/first_network_device.c
+++ b/src/first_network_device.c
@@ -6,11 +6,12 @@
 
 #include "i3status.h"
 
-const char *first_eth_interface(const net_type_t type) {
+const char *first_eth_interface(const net_type_t type, const char *title) {
     static char *interface = NULL;
     struct ifaddrs *ifaddr, *addrp;
     struct stat stbuf;
     static char path[1024];
+    static char first[1024];
 
     getifaddrs(&ifaddr);
 
@@ -31,11 +32,16 @@ const char *first_eth_interface(const net_type_t type) {
         if (addrp->ifa_addr->sa_family != AF_INET &&
             addrp->ifa_addr->sa_family != AF_INET6)
             continue;
-        // Skip this interface if it is a wireless interface.
+        // Skip this interface if it is not the right type of interface.
         snprintf(path, sizeof(path), "/sys/class/net/%s/wireless", addrp->ifa_name);
         const bool is_wireless = (stat(path, &stbuf) == 0);
         if ((is_wireless && type == NET_TYPE_ETHERNET) ||
             (!is_wireless && type == NET_TYPE_WIRELESS))
+            continue;
+        // Skip this interface if the interface prepended with _first_ does
+        // not match the first strlen(title) characters of title.
+        snprintf(first, sizeof(first), "_first_%s", addrp->ifa_name);
+        if (strncasecmp(title, first, strlen(title)))
             continue;
         interface = strdup(addrp->ifa_name);
         break;
--- a/i3status.c
+++ b/i3status.c
@@ -642,9 +642,7 @@ int main(int argc, char *argv[]) {
 
             CASE_SEC_TITLE("wireless") {
                 SEC_OPEN_MAP("wireless");
-                const char *interface = NULL;
-                if (strcasecmp(title, "_first_") == 0)
-                    interface = first_eth_interface(NET_TYPE_WIRELESS);
+                const char *interface = first_eth_interface(NET_TYPE_WIRELESS, title);
                 if (interface == NULL)
                     interface = title;
                 print_wireless_info(json_gen, buffer, interface, cfg_getstr(sec, "format_up"), cfg_getstr(sec, "format_down"));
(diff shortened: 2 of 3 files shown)


src/first_network_device.c
```diff
@@ -31,12 +32,17 @@ const char *first_eth_interface(const net_type_t type) {
...
-        // Skip this interface if it is a wireless interface.
+        // Skip this interface if it is not the right type of interface.
         snprintf(path, sizeof(path), "/sys/class/net/%s/wireless", addrp->ifa_name);
         const bool is_wireless = (stat(path, &stbuf) == 0);
         if ((is_wireless && type == NET_TYPE_ETHERNET) ||
             (!is_wireless && type == NET_TYPE_WIRELESS))
             continue;
+        // Skip this interface if the interface prepended with _first_ does
+        // not match the first strlen(title) characters of title.
+        snprintf(first, sizeof(first), "_first_%s", addrp->ifa_name);
```
Georgiy-Tugai: You probably want `snprintf(first, sizeof(first)-1`... in order to leave space for the NUL character.

howlett: I don't think I do.  From the man page "The functions snprintf() and  vsnprintf()  write  at  most  size  bytes (including the terminating null byte ('\0')) to str."  Is this an issue with using another snprintf implementation?  If so, then does line 35/36 need to be fixed as well?

Georgiy-Tugai: My apologies, I misremembered the specification.

I am not aware of any snprintf implementations which do not match the
behaviour your code relies upon.

stapelberg:
This pull request doesn’t sound like it’s adding behavior which would be generally useful to and necessary for most of our users:

* since udev/systemd v197, interface names [should be stable](https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/), so the behavior you’re describing sounds like a bug in and of itself
* elsewhere in i3status, we don’t have fuzzy/prefix/regexp matching, so this introduces a special case

I’d suggest:

1. For a low-effort fix, generate `i3status.conf` on boot from a template with the proper device name.
2. Figure out why the network device name isn’t stable, get that fixed.

In case point 2 isn’t feasible and affects more than just your specific hardware combination, I’d be willing to reconsider this functionality.

src/first_network_device.c
howlett: No worries, I'm happy you took time to review.  I'm not familiar with clang which is included in the testing so that's why I asked about other implementations.

howlett:
Thank you for your reply but I believe that I was not clear as to why my patch is needed.

1. Android phones create a random hardware address for tethering on each reboot of the phone (not the PC reboot).  This is not a bug, it's by design.
2. udev follows the correctly documented procedure to create the device name from your link which reaches option #4: "Names incorporating the interfaces's MAC address (example: enx78e7d1ea46da)"
3. Since the device name changes on Phone reboots, the i3status.conf will not find the correct name even if generated on PC reboots.

Admittedly, this could be fixed by creating my own custom udev rule that would create a link to the right device in /dev.  However, adding a udev rule would require root privileges to make this change and is obviously not as portable as updating a git repo of config files to get the correct i3status.conf.

stapelberg:
Thanks for the additional context.

Commit 15de209cbaee6827f061e8d2788c9db232012c6b starts looking at the DEVTYPE= property. Given that you use prefix matching for effectively selecting an interface by type, could the same outcome be achieved by looking at DEVTYPE= or another field in the uevent file?

howlett:
No, that change does not work because the tethered interface is considered an ethernet device.  I pulled the new code and tested.  My tethered phone shows up as the first ethernet device when eth0 is down.

stapelberg:
Can you provide the full uevent file for the device in question please?

howlett:
I don't see a DEVTYPE in either of my wired device uevent files, but I have included all three files at the end of this message.

I added debug to the new code to see what enx* type was:
ifa_name = wlan0
wlan0:
type = 1 iftype = 0
ifa_name = enx7a620ffa3440
enx7a620ffa3440:
type = 1 iftype = 1
done

This is the same type that is returned by eth0 when it is active:
ifa_name = eth0
eth0:
type = 1 iftype = 1
done


Here are the files:
[jedix@wolf 10:07:32 i3status]$ cat /sys/class/net/eth0/uevent 
INTERFACE=eth0
IFINDEX=2
[jedix@wolf 10:07:34 i3status]$ cat /sys/class/net/enx7a620ffa3440/uevent 
INTERFACE=enx7a620ffa3440
IFINDEX=9
[jedix@wolf 10:07:40 i3status]$ cat /sys/class/net/wlan0/uevent                                                                       
DEVTYPE=wlan                                                                                                                          
INTERFACE=wlan0                                                                                                                       
IFINDEX=3

stapelberg:
Thanks for posting the additional details.

I looked around a bit:

There’s a ubuntu bug report for this issue at https://bugs.launchpad.net/ubuntu/+source/dbus-property-service/+bug/1570800

Based on what the reporter writes, I think it makes sense to tackle this issue higher up in the stack, i.e. in udev, where you can assign a static interface name based on other attributes, such as the USB serial number of the device. Maybe http://www.monblocnotes.com/node/1895 helps to come up with such a udev configuration. I don’t have an Android phone here, so I can’t provide a working example unfortunately.

The rationale for addressing the issue in udev instead of i3status is that the approach in this pull request (matching on the interface name) feels like a workaround to me — it only works because you have precisely one device on your machine which uses MAC-address based naming. The approach won’t work for users who have multiple MAC-address based named devices. Notably, I found reports of Raspberry Pi users, whose ethernet device already uses MAC-address based naming.

Please also update the following questions once you’ve found a working solution:
https://askubuntu.com/questions/721413/how-do-i-get-usb-tethering-working-between-15-10-and-an-android-device
https://wiki.archlinux.org/index.php/Talk:Android_tethering

Thanks!

PR #158, opened 2016-08-19
Added missing ampere to watt conversion for last full capacity.

### LINKED ISSUE #157: Incorrect battery values for users using last full capacity with ampere to watt conversion.
After the refactoring made in d74bc90bfe8ac11ba3d6ac5326d95dedd10c39c3, the logic behind last full capacity calculation changed, breaking it for users that who were also using ampere to watt conversion.

Before the change, there was only `full_design` field. Its value was assigned depending on the `last_full_capacity` option - it was either full design or last full capacity reported. If a ampere to watt conversion was required, `full_design` was correctly converted.

However, since after the change, there are two separate fields for full design capacity (`full_design`) and last full capacity (`full_last`). Unfortunately, if an ampere to watt conversion is required, only the `full_design` gets converted. If we try to use `full_last`, the units don't match and we get incorrect results (way over 100%, which get capped at 100% battery status).

tnnn:
This change fixes #157 by adding missing conversion.


tommie:
+1, my bad for omitting this in the refactoring.

Issue #126, opened 2016-05-08
volume: able to specify pulse device via "pulse[:N]" or "pulse[:Name]"


ntzrmtthihu777:
Shouldn't be terribly hard, but to be frank I lack the c chops to do it in a way that isn't ugly as hell.

Reason: The device index is variable, whereas the device name appears to be pretty constant even across reboots and unplug/replugs.

Proof: run `pacmd list-sinks| grep -e 'card:' -e 'device.id' -e 'index:' -e 'alsa.card_name'`, unplug and replug a device and re-run it. Each time the device will increment its index.

Watcom:
You can specify the default sink in /etc/pulse/default.pa with a line like this:
`set-default-sink alsa_output.usb-Creative_Technology_Ltd_SB_X-Fi_Surround_5.1_Pro_00000IWp-00-Pro.analog-stereo`
and then leave the device as empty or "default" in i3status.conf. Then you can unplug/replug and the device and it *should* stick with it (I haven't tried it though).

ntzrmtthihu777:
Oh I know about that, do it a fair amount of time, but suppose one wants multiple displays of volume for multiple sound devices? Can't very well have two defaults ;)
Being able to specify an exact device by name would be very useful.

Watcom:
Yes, that's true. I have to say it makes me cringe a little to force PulseAudio to resolve the device name once per second, so it would have to do it at startup. But then you wouldn't be able to follow the device dynamically as you replug it without restarting i3status.

ntzrmtthihu777:
oh yeah. didn't think about that portion of it... hrm. perhaps a periodic refresh of name->index mappings outside of the 1s loop?

Watcom:
Ideally it should subscribe to PulseAudio events so that it detects a change and only then resolve any new indices.
I'm curious about how useful this would be, though. Do you have such a setup, I mean with multiple sound cards and you actually need to see the volume of more than one simultaneously? As opposed to using one at a time and having PulseAudio decide which is the default based on what's plugged in?

ntzrmtthihu777:
I have two main sound cards, the one on my GPU and a Logitech G933; I keep the GPU at max volume and control the actual volume on the tv I use as a display, so I'm really only interested in the headset at any given time, but I do believe this could be useful for folks with more advanced audio setups.

(I also have an on-mobo card, but I never use it, so its moot to this discussion)

Watcom:
See, this issue has so little demand not even the requester needs it! ;)
[...]

PR #162, opened 2016-09-12
Pulse device may be specified by name (#126)

Watcom:
Implements #126 
Using the default sink (by not specifying a sink) still works as before, i.e., if you change the default sink from device A to device B it changes the volume displayed to device B. On any other situation (if you specify the sink by id or name), it should follow that device even if it changes sink id. Of course if you specify by id it won't survive an i3status restart, so specifying the sink by name is better.

### DIFF
--- a/man/i3status.man
+++ b/man/i3status.man
@@ -487,10 +487,16 @@ or
 
  device = "pulse:N"
 
-where N is the index of the PulseAudio sink. If no sink is specified the
-default is used. If the device string is missing or is set to "default",
-PulseAudio will be tried if detected and will fallback to ALSA (Linux)
-or OSS (FreeBSD/OpenBSD).
+where N is the index or name of the PulseAudio sink. You can obtain the name of
+the sink with the following command:
+
+ $ pacmd list-sinks | grep name:
+            name: <alsa_output.pci-0000_00_14.2.analog-stereo>
+
+The name is what's inside the angle brackets, not including them. If no sink is
+specified the default sink is used. If the device string is missing or is set
+to "default", PulseAudio will be tried if detected and will fallback to ALSA
+(Linux) or OSS (FreeBSD/OpenBSD).
 
 *Example order*: +volume master+
 
@@ -514,6 +520,13 @@ volume master {
 	format = "♪: %volume"
 	format_muted = "♪: muted (%volume)"
 	device = "pulse:1"
+}
+-------------------------------------------------------------
+-------------------------------------------------------------
+volume master {
+	format = "♪: %volume"
+	format_muted = "♪: muted (%volume)"
+	device = "pulse:alsa_output.pci-0000_00_14.2.analog-stereo"
 }
 -------------------------------------------------------------
 
--- a/src/print_volume.c
+++ b/src/print_volume.c
@@ -4,6 +4,7 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <err.h>
+#include <ctype.h>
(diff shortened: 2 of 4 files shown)


src/pulse.c
```diff
@@ -88,20 +95,22 @@ static void store_volume_from_sink_cb(pa_context *c,
...
 
-static void get_sink_info(pa_context *c, uint32_t idx) {
+static void get_sink_info(pa_context *c, uint32_t idx, const char *name) {
     pa_operation *o =
-        idx == DEFAULT_SINK_INDEX ? pa_context_get_sink_info_by_name(
-                                        c, "@DEFAULT_SINK@", store_volume_from_sink_cb, NULL)
-                                  : pa_context_get_sink_info_by_index(
-                                        c, idx, store_volume_from_sink_cb, NULL);
-    pulseaudio_free_operation(c, o);
+        name || idx == DEFAULT_SINK_INDEX ? pa_context_get_sink_info_by_name(
```
stapelberg: Could you refactor this nested ternary operator conditional into a couple of if statements please? I think that’d be easier to read bottom line.

src/pulse.c
```diff
@@ -183,14 +192,14 @@ static void context_state_callback(pa_context *c, void *userdata) {
...
-int volume_pulseaudio(uint32_t sink_idx) {
+int volume_pulseaudio(uint32_t sink_idx, const char *sink_name) {
     if (!context_ready || default_sink_idx == DEFAULT_SINK_INDEX)
         return -1;
 
     pthread_mutex_lock(&pulse_mutex);
     const indexed_volume_t *entry;
     TAILQ_FOREACH(entry, &cached_volume, entries) {
-        if (entry->idx == sink_idx) {
+        if (sink_name ? entry->name && !strcmp(entry->name, sink_name) : entry->idx == sink_idx) {
```
stapelberg: Same here, can you avoid the ternary please?

src/print_volume.c
```diff
@@ -69,7 +70,8 @@ void print_volume(yajl_gen json_gen, char *buffer, const char *fmt, const char *
     if (!strncasecmp(device, "pulse", strlen("pulse"))) {
         uint32_t sink_idx = device[5] == ':' ? (uint32_t)atoi(device + 6)
                                              : DEFAULT_SINK_INDEX;
-        int cvolume = pulse_initialize() ? volume_pulseaudio(sink_idx) : 0;
+        const char *sink_name = device[5] == ':' && !isdigit(device[6]) ? device + 6 : NULL;
```
stapelberg: Didn’t spot this in the previous reviews, but could you switch from magic numbers to e.g. `strlen("pulse:")`? The compiler will translate these at compile-time.

man/i3status.man
```diff
@@ -487,7 +487,7 @@ or
 
  device = "pulse:N"
 
-where N is the index of the PulseAudio sink. If no sink is specified the
+where N is the index or name of the PulseAudio sink. If no sink is specified the
```
stapelberg: Could you add an example here, and refer people to where they can find out about the sinks on their system? E.g., `pactl list sinks`, IIUC.

stapelberg:
Looks good overall.

Watcom:
Done, commit pushed.

Watcom:
Actually I just spotted a possible problem with the logic, it could incorrectly match by index if two devices are specified and only one by name. Let me fix this first.

Watcom:
OK, done.

PR #163, opened 2016-09-17
Use glob to get all batteries

zhaostu:
It seems that on many laptops BAT1 is present instead of BAT0. The change takes care of this very common case.

### DIFF
--- a/src/print_battery_info.c
+++ b/src/print_battery_info.c
@@ -11,7 +11,7 @@
 
 #if defined(LINUX)
 #include <errno.h>
-#include <sys/stat.h>
+#include <glob.h>
 #include <sys/types.h>
 #endif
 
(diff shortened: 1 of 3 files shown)


zhaostu:
Looks like the Linux kernel only supports 4 batteries at this point: https://github.com/torvalds/linux/blob/9256d5a308c95a50c6e85d682492ae1f86a70f9b/drivers/acpi/sbs.c#L622

And the ACPI driver may skip generating a sysfs entry depending on the "batteries_supported" bit. I am not sure about other operating systems though.

zhaostu:
@stapelberg any thoughts on this one?

eplanet:
If I may share my thoughts, I think the idea of your PR is good, however I'd have used [`glob`](https://linux.die.net/man/3/glob) instead of some condition hack. This way we could get rid of the [1 to 1000 loop](https://github.com/zhaostu/i3status/blob/62e802b8189015f3223f50824df8f0425c594e17/src/print_battery_info.c#L418), with an appropriate way of discovering all available batteries.

It would both solve the problem of a battery starting with 1 instead of 0, and discover all available batteries without assuming some maximum amount of batteries.

zhaostu:
@eplanet Make sense, I was just trying to modify the code as little as possible, not knowing why the decision was made in the first place. I'll come up with a change when I've got time.

eplanet:
That actually isn't for me to decide, since @stapelberg will eventually choose to merge or not your PR. You should wait for his validation prior to dive into any unnecessary coding :-)

zhaostu:
@stapelberg haven't heard from you. I would like to hear what you think. Thanks.

Airblader:
Michael is quite busy with work at time. He'll look at this, but it might just take a few days. :-)

stapelberg:
Thanks for your patience with my reply.

I think the idea of improving the error path makes sense. I agree with @eplanet that using `glob()` seems like the best solution here in order to quickly detect any available batteries. We should pay attention to using a sorted list here so that the code is deterministic.

zhaostu:
@stapelberg thanks for the reply. I'll do some research and modifications, and update this pull request later.

eplanet:
Why did you close the PR? Tell me if you don't work on it so I can implement the fix.

zhaostu:
@eplanet @stapelberg Ah, I didn't mean to, I force pushed master to clean up my master branch and Github think there is no code difference and closed this pull request.

But when trying to implement the glob, I was facing some issues :confused: . The configuration for "path" for battery is encoded as `/sys/class/power_supply/BAT%d/uevent`, but in order to do glob, `*` has to be in `%d`'s place. Since C does't have built in string replacement, it can get messy. So I am thinking of three solutions:

1. Just loop from 1 to 4, since we know Linux kernel only supports 4, and the code is wrapped in "#ifdef(LINUX)", so we are safe here. This is the simplest solution I can think of.
2. Change `%d` to `%s` in the configuration, in order to do a `snprintf` on that string to replace `%s` to `*`. But this may break people's configurations.
3. Write a string replacement function myself, which gets pretty messy.

Please let me know what you all think, or have any better solutions. Thanks!

Airblader:
I don't see an issue with implementing string replacement. It's a standard task with many resources. 

And at least in libi3 we already have format_placeholders:

https://github.com/i3/i3/blob/next/libi3/format_placeholders.c

zhaostu:
@Airblader is there a way to reuse that code? Or should I basically copy that into i3status.

Airblader:
I don't think there currently is a way to reuse it. libi3 isn't available to i3status.

zhaostu:
I basically copied some code from libi3 and used that to replace "%d" to "*", and used glob to find files matching the pattern.

src/print_battery_info.c
```diff
@@ -414,40 +414,36 @@ static bool slurp_all_batteries(struct battery_info *batt_info, yajl_gen json_ge
...
+    if (!strcmp(globpath, path)) {
+        OUTPUT_FULL_TEXT("no '%d' in battery path");
+        return false;
+    }
 
-            OUTPUT_FULL_TEXT(format_down);
-            return false;
+    glob_t globbuf;
+    int res = glob(globpath, 0, NULL, &globbuf);
+    if (res == 0) {
```
stapelberg: You don’t seem to use `res` more than once, so you can simplify this to:

```c
if (glob(globpath, 0, NULL, &globbuf) == 0) {
```

src/print_battery_info.c
```diff
@@ -414,40 +414,36 @@ static bool slurp_all_batteries(struct battery_info *batt_info, yajl_gen json_ge
...
-        }
-
-        /* Probe to see if there is such a battery. */
-        struct stat sb;
-        if (stat(batpath, &sb) != 0) {
-            /* No such file, then we are done, assuming sysfs files have sequential numbers. */
-            if (errno == ENOENT)
-                break;
+    placeholder_t placeholders[] = {
+        {.name = "%d", .value = "*"},
```
stapelberg: `format_placeholders` is useful when the replacement values are dynamic (i.e. not known at compile-time) and longer than the bit they are replacing. Modifying a string to replace all `%d` values with `*` can easily be done without `format_placeholders`:

```c
        char *placeholder;
        while ((placeholder = strstr(battery, "%d")) != NULL) {
                *placeholder = '*';
                // strings passed to strcpy() must not overlap:
                char *tmp = strdup(placeholder+2);
                strcpy(placeholder+1, tmp);
                free(tmp);
        }
```

Can you update the code accordingly (and thereby reduce the complexity of this PR significantly) please?

zhaostu:
@stapelberg I changed the code based on your feedback.

I didn't implement the string replacement exactly as how you put. Specifically, I didn't use a for loop, because in the man page, it specifically indicated the [first occurrence of %d]( https://github.com/i3/i3status/blob/master/man/i3status.man#L331) will get replaced. So I don't think the while loop is necessary.

stapelberg:
The changes to src/print_battery.c look good, but could you reverse the unrelated changes that move functions into safewrappers.c? I’d like to keep pull requests as small as possible, and that change seems to be unnecessary at this point.

zhaostu:
@stapelberg I need to use the `sstrdup` function already defined in the i3status.c. If I just add the function into i3status.h will it work?

zhaostu:
@stapelberg updated

Issue #166, opened 2016-09-28
Remove wireless attribute padding?


jbernard:
Hallo, I removed the right-justification / padding in the wireless attributes such that leading 0's are no longer printed and I find the result to be quite pleasing.  I suspect there is a reason for not doing this earlier (consistent widths perhaps) but I wanted to suggest the change anyway just in case:

https://github.com/jbernard/i3status/commit/4d21a18599538292eb29acd3189c354e21acd2da

If there is interest, I'm happy to rebase and submit a pull request.

stapelberg:
The `%03d` format string was introduced in commit 71f71660322b12b8602f5979222d280b7db333bd in 2010. The rationale is to have the width fixed, so that the different status modules don’t jump around when your wireless signal changes.

Entirely removing the padding does not seem like something we’d want to do, but using `%3d` instead of `%03d` (i.e. blank-padded instead of zero-padded) should be fine. If you’re interested in that, please send a pull request.

devkev:
Should this be closed, since the fix was merged in #167?

Airblader:
Yes, thanks!

PR #167, opened 2016-10-03
Use blank padding for wireless quality (#166)

jbernard:
This patch changes the existing zero-padded wireless quality info to use
blank padding with a width of two instead of three.  A 100% wireless
signal quality is difficult to achieve even in the most ideal
conditions and I think this improves readability by a significant
margin.

### DIFF
--- a/src/print_wireless_info.c
+++ b/src/print_wireless_info.c
@@ -497,7 +497,7 @@ void print_wireless_info(yajl_gen json_gen, char *buffer, const char *interface,
         if (BEGINS_WITH(walk + 1, "quality")) {
             if (info.flags & WIRELESS_INFO_FLAG_HAS_QUALITY) {
                 if (info.quality_max)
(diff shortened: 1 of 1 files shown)


jbernard:
I left signal and noise level out because I don't use them personally and I'm not sure how others feel about it.  But i'm happy to add them back if needed.

stapelberg:
> A 100% wireless
> signal quality is difficult to achieve even in the most ideal
> conditions

…but drivers still report it occasionally, at least on some machines. Please go back to 3 digits of padding.

> I left signal and noise level out because I don't use them personally and I'm not sure how others feel about it. But i'm happy to add them back if needed.

Yes, please make your change consistent.

Airblader:
@stapelberg Seems like there's been an update since your last comment.
@jbernard When you make changes please leave a comment as we won't get notified otherwise.

PR #168, opened 2016-10-23
Allow setting custom locale in tztime configuration

bkhl:


stapelberg:
What’s the use-case for this? I.e., why not just set the preferred locale for i3status itself?

Also, please squash your commits into logical entities so that our git history stays clean.

bkhl:
I wanted to show my birth country's time zone in that country's locale, and my local time in my current locale.

stapelberg:
> I wanted to show my birth country's time zone in that country's locale, and my local time in my current locale.

Can you update the commit description to explain this as the motivation please?

src/print_time.c
```diff
@@ -45,6 +83,8 @@ void print_time(yajl_gen json_gen, char *buffer, const char *title, const char *
     set_timezone(tz);
     localtime_r(&t, &tm);
 
+    set_locale(locale);
```
bkhl: You are right, I was under the misapprehension that the program needs to reset to locale, as if it was actually working by changing the environment variables.

bkhl:
Have updated the commit description, as well as the code according to the review above.

src/print_time.c
stapelberg: Can you get rid of the `set_locale` function entirely please? I don’t think it helps in understanding what the program does.

I would have expected the following code:
```c
if (locale != NULL) {
  setlocale(LC_ALL, locale);
}
// …rest of print_time() code…
if (locale != NULL) {
  setlocale(LC_ALL, "");
}
```

…so that the locale is reset before other modules are run.

bkhl: That's reasonable, done.

PR #170, opened 2016-10-23
Unit testing in Travis

eplanet:
Here's an attempt to add some unit testing to i3status. It might not be the perfect methodology, but it works fine and could lead the path to a lot more testing.

@stapelberg @Airblader, could you share your thoughts before I go further in that direction? Thank you very much in advance.


man/i3status.man
```diff
@@ -22,6 +22,9 @@ configuration files in the following order:
 3. ~/.i3status.conf
 4. /etc/i3status.conf
 
+-o::
+Specifies i3status to run only once before quitting for testing purposes.
```
Airblader: Test flags shouldn't be documented because we don't want anybody using them. Otherwise we can never change them.

travis/run-tests.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+TESTDISK=`./i3status -o -c test/print_disk_info.conf`
```
Airblader: The problem with this – and this really is the single biggest challenge with introducing tests for i3status – is that it relies on the environment on which the tests are run. And to me that's a major issue.

Perhaps the way to go is to run tests within a virtualized environment (much like the i3 tests usually run in a nested X server.) Admittedly this increases the effort significantly, but tests relying on the executing user's environment just lead to too much headache.

travis/print_disk_info.conf
```diff
@@ -0,0 +1,8 @@
+general { output_format="xmobar" }
```
Airblader: This should be properly formatted. I also don't think xmobar should be the output format for our tests – the vast majority  of users will use it with i3bar, which is also the i3-internal usecase, so that's what we should test, IMHO. That'll require reading JSON, but I think we need to invest that effort.

man/i3status.man
eplanet: I understand the point, but I'd like it to be exposed since you can definitely want to test various configuration with a "run once" mode, even from a user point of view. You confirm it still shouldn't be documented?

travis/run-tests.sh
eplanet: That definitely is the biggest challenge with `i3status` (and `i3lock` by the way)… However the disk test is only assuming a mount point exists for the Linux root '/', that definitely is a mount point that must exist on any Linux environment, be it virtual or containerized. Don't you agree?

But the challenge will be tremendous for each other use case…

man/i3status.man
Airblader: Yes, I still think it shouldn't be exposed. This PR is about tests, not enhancing anything for the user. If we want to expose this later on we can still do so. But we should do that if anybody actually has a need to do so.

travis/run-tests.sh
Airblader: I agree that it might work for this case, but even so I think it's wrong from a theoretical point of view. I don't think having a testsuite is much use if we cannot test anything reliably across machines.

stapelberg: Agreed. The most useful feature of a test suite for i3status is to simulate different environments (e.g. users report a battery-related bug, we add a testcase which gives i3status precisely the values which the bug reporter is seeing).

Battery is a low-hanging fruit, because it merely requires reading a different text file, other APIs are way harder (e.g. cfg80211 for wifi — I don’t expect tests for that API, at least not in the beginning).

travis/print_disk_info.conf
eplanet: I'd like to go for "none" at the beginning. Like you said, setting up a clean test environment is the priority. :-)

travis/run-tests.sh
eplanet: I added a test case for battery using values found on my laptop. Do you think I should remove the test case for disk? I kinda liked it… :-)

man/i3status.man
eplanet: OK. Removed.

travis/run-tests.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
stapelberg: Let’s not build this in shell — building anything non-trivial in shell is an exercise in frustration.

Can we use Perl, much like the i3 testsuite itself? (For consistency and because i3 contributors are already familiar with it.)

travis/print_battery_info.conf
```diff
@@ -0,0 +1,10 @@
+general {
```
stapelberg: Here and below, with an eye towards future growth, I think it’d be good to move all test files into a directory named after the test, e.g.:

```
testcases/
testcases/001-battery/
testcases/001-battery/i3status.conf
testcases/001-battery/BAT0_uevent
testcases/001-battery/expected_output.sh
```

…where expected_output.txt (to be read) or expected_output.sh (to be executed) are compared against the actual i3status output automatically.

If necessary, we can also have tests override the test behavior, i.e. doing something else than just comparing text, by placing another file in their directory which is then sourced and executed.

Also note that the 1-test-per-directory model enables strong parallelism from the very beginning.

i3status.c
```diff
@@ -505,7 +506,7 @@ int main(int argc, char *argv[]) {
     if (setlocale(LC_ALL, "") == NULL)
         die("Could not set locale. Please make sure all your LC_* / LANG settings are correct.");
 
-    while ((o = getopt_long(argc, argv, "c:hv", long_options, &option_index)) != -1)
+    while ((o = getopt_long(argc, argv, "c:hvo", long_options, &option_index)) != -1)
```
stapelberg: In general, I’d prefer it if options which are not supposed to be used by users directly (such as the run-once option) do not get a short character, but only a long form (i.e. are added to `long_options`). That way, we don’t “waste” short characters, and we make it a bit less likely for people to use these options by accident.

travis/run-tests.sh
eplanet: I actually happen to prefer bash over perl myself, especially for unit testing, but if that is consistent with the rest of the contributions, that's fair enough.

eplanet:
First, thanks for you enthusiasm over this PR :nerd_face:. I think we have a good candidate for unit testing from now.

eplanet:
Could you confirm that it is OK, or give me comments on what to change to comply? Thanks!

stapelberg:
Conceptually, I think this looks like a good start. Do you want an actual review & merge, or is there still something you want to change?

eplanet:
Well there's obviously so much more to do in unit testing, so I'd prefer the "bootstrap" to be merged, in order to have other people contribute and unit testing embedded in Travis right away. Maybe I'm too impatient?

In any case, I sure will try to do more test cases, but in separate PR, if that's OK for you?

i3status.c
```diff
@@ -483,6 +484,7 @@ int main(int argc, char *argv[]) {
         {"config", required_argument, 0, 'c'},
         {"help", no_argument, 0, 'h'},
         {"version", no_argument, 0, 'v'},
+        {"run-once", no_argument, 0, 1},
```
stapelberg: To be consistent with the i3 source code, please use `0` instead of `1` as last parameter, then in the `getopt_long` loop,  use strcmp on long_options[option_index].name

travis/run-tests.pl
```diff
@@ -0,0 +1,42 @@
...
+        print BOLD, RED, "Failed!\n", RESET;
+        return 0;
+    }
+}
+
+
+my $testcases = 'testcases';
+my $testresults = 1;
+
+opendir(DIR, $testcases) or die $!;
```
stapelberg: Instead of named handles (`DIR`), please use variables:

```perl
opendir(my $dir, $testcases) or die "Could not open directory $testcases: $!";
```

travis/run-tests.pl
```diff
@@ -0,0 +1,42 @@
...
+
+while (my $file = readdir(DIR)) {
+    next unless (-d "$testcases/$file");
+    next if ($file =~ m/^\./);
+    print "Testing test case '$file'… ";
```
stapelberg: Use `say` instead of `print` to get a newline. Requires adding `use v5.10;` to the preamble.

travis/run-tests.pl
```diff
@@ -0,0 +1,42 @@
...
+    }
+}
+
+
+my $testcases = 'testcases';
+my $testresults = 1;
+
+opendir(DIR, $testcases) or die $!;
+
+while (my $file = readdir(DIR)) {
```
stapelberg: nit: please rename `$file` to `$entry`. Yes, strictly speaking a directory is also just a file, but given that `readdir` returns directory entries, I think `$entry` is more appropriate :).

travis/run-tests.pl
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env perl
+
+use strict;
+use warnings;
+use Term::ANSIColor qw(:constants);
+
+sub TestCase {
+    my $n = scalar(@_);
```
stapelberg: `$n` is unused? Remove?

travis/run-tests.pl
```diff
@@ -0,0 +1,42 @@
...
+    my $conf = "@_/i3status.conf";
```
stapelberg: Instead of using `@_` directly here, please capture subroutine arguments into a named variable:

```perl
  my ($dir) = @_;
  my $conf = "$dir/i3status.conf";
  # …
```

travis/run-tests.pl
```diff
@@ -0,0 +1,42 @@
...
+    my $testres = `./i3status --run-once -c $conf`;
+    my $refres = "";
+
+    if ( -f "@_/expected_output.txt") {
+        $refres = `cat "@_/expected_output.txt"`;
+    } elsif ( -f "@_/expected_output.sh") {
+        $refres = `bash @_/expected_output.sh`;
+
```
stapelberg: nit: remove empty line

i3status.c
eplanet: I might be going further than expected, but I changed the whole getopt_long loop to a `switch` case, exactly like i3. I thought it would be even more consistent that way.

travis/run-tests.pl
eplanet: Indeed… :-)

travis/run-tests.pl
eplanet: Please forgive my perl… :-)

travis/run-tests.pl
eplanet: It makes sense :-)

travis/run-tests.pl
eplanet: Doing so will break the display originally intented: I thought it would be nice when running the test cases to get one line for each with its name and its colored result ("**OK**" or "**Failed!**") on the same line, for easy summarization. Using `say` will print a new line without the colored result.

I suggest I change that behavior to still print on the same line each test case name with its colored result.

eplanet:
I made the requested changes. Please share your thoughts if anything is wrong.

Issue #177, opened 2016-11-13
Provide format_bellow_threshold and format_above_threshold for disk monitor


gfa:
Hello

I'd like to have the possibility to configure different format for disk output depending if the disk is above or bellow threshold, my goal here is not to have output when the free space is above but a big fat red warning when the free space is bellow threshold.

I do the same for my nic in this way

ethernet enp0s25 {
        # if you use %speed, i3status requires root privileges
        format_up = "E: %ip (%speed)"
        format_down = ""
}

when it is down (normal) it doesn't print anything, but when the nic is in use it prints its status

thanks!

stapelberg:
This makes sense to me, pull requests welcome.

We probably should implement such an option for all modules which offer configurable thresholds.

eplanet:
@gfa if you're not willing to work on this, I'd be happy to come up with a PR.

gfa:
@eplanet please do, thanks

eplanet:
Should be closed since #179.

PR #179, opened 2016-11-14
Provide format_above_threshold/format_below_threshold options

mihaicmn:
See #177 for more details. Any feedback is appreciated.


Airblader:
I'd like if we can take this opportunity to add tests for this, given that we just introduced some infrastructure for that.

mihaicmn:
Agreed! Should I use this PR or create a new one?

Airblader:
Change and tests go together, so this PR (which I haven't looked at yet).

eplanet:
@mihaicmn while it's great that you worked on this, you **should** have mentioned it in the issue thread, since somebody else can be working in parallel on the same issue without knowing it's a waste of time…

mihaicmn:
@eplanet You are right, I apologize for the inconvenience. From now on I will definitely keep in mind your advice.

PR #181, opened 2016-11-18
Add brightness indicator

mihaicmn:
I needed a module to display the level of brightness then I thought others might be interested in using it.

This new indicator requires a configuration section as below:
order += "brightness"
brightness {
        format = "◐: %brightness"
        actual_brightness_path = "/sys/class/backlight/intel_backlight/actual_brightness"
        max_brightness_path = "/sys/class/backlight/intel_backlight/max_brightness"
}

If you have any suggestions/comments on this please let me know.

PR #184, opened 2016-12-03
Add Linux display of backlight intensity level

adrieng:
The percentage is computed by reading the "backlight" and "max_backlight" files
in /sys/class/backlight/DEVICE.

### DIFF
--- a/i3status.c
+++ b/i3status.c
@@ -465,6 +465,16 @@ int main(int argc, char *argv[]) {
         CFG_CUSTOM_SEP_BLOCK_WIDTH_OPT,
         CFG_END()};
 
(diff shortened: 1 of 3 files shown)


stapelberg:
#38, #109, #73, #181

adrieng:
Sorry, I didn't see the previous PRs.

However, doesn't the argument "You can already *see* brightness, just make your screen brighter/dimmer if you're not satisfied" also apply mutatis mutandis to the volume indicator already present in i3? "You can already *hear* sound, just make sound louder/quieter if you're not satisfied."

stapelberg:
You have a point. However, given that sound is already included, we won’t take it out, as we try hard not to break existing setups.

PR #188, opened 2016-12-23
Consolidate measure speed script

mwarning:
Some cleanup of old net-speed scripts.

Changes to net-speed script:
- add enp* wlp* network interfaces (naming scheme on Arch Linux)
- store time in variable instead of parsing the i3 config file
- add .sh suffix to file

.. let me know if changes need to be done


mwarning:
btw., I got the ok from @azuwis to change the license to Public Domain.

Airblader:
Public domain is not a license. We should reject this (license) change, it's quite problematic.

Airblader:
Even the original file stated public domain. If this license is kept I vote for removing this script from the repository and let somebody else maintain it elsewhere. We shouldn't keep multiple licenses in the i3 suite at all, but in particular no license is a real issue.

mwarning:
Since both parties agree to place it into the public domain, we can give it a license that fits i3status.

If you want to remove the scripts, feel free to do so. But even so, I have seen the network speed information on almost all i3 installations.

mwarning:
@Airblader is it ok to just remove the license line or do we have to add the i3status license text?

azuwis:
I've made a pull request to change the license of contrib/netspeed to the same license as i3status itself, see https://github.com/i3/i3status/pull/189.

mwarning:
@azuwis ok, I've added the license line to the new script. I hope that will do it.

stapelberg:
@breunigs Are you okay with this change? I’m not familiar with these scripts.

breunigs:
I like the general idea, because this might fix the bug where it shows ridiculous speeds after unlocking your computer, because of the non-adjusted time interval.

As it is however, the script is prone to division by zero when the time interval is short (e.g. through a binding that calls `killall -SIGUSR1 i3status`, say on volume change). I guess switching to `date +%s%N` might work, though I haven't tried.

breunigs:
Also I don't get why you removed the examples on how to manually configure the desired interfaces.

mwarning:
@breunigs: thanks for the hint. I have added a check > 0 and the example lines were added again.
Do you want to be added to the author list?

breunigs:
No I'm good

mwarning:
The merge conflict was solved and one possible error case fixed

Imho, it is ready to be merged.

PR #195, opened 2017-01-30
Fix the deciKelvin to Celsius conversion

bapt:


src/print_cpu_temperature.c
```diff
@@ -13,7 +13,7 @@
 #include <err.h>
 #include <sys/types.h>
 #include <sys/sysctl.h>
-#define TZ_ZEROC 2732
+#define TZ_ZEROC 2731
```
stapelberg: [Wikipedia](https://en.wikipedia.org/wiki/Kelvin) states:

> The definition implies that absolute zero (0 K) is equivalent to −273.15 °C (−459.67 °F).

It seems like previously, we used the [ceil()](https://manpages.debian.org/ceil.3)ed value, and you’re suggesting to use the [floor()](https://manpages.debian.org/floor.3)ed value. Can you outline (also in your commit description) why this change makes things better please?

bapt: The main reason is a followup on that https://svnweb.freebsd.org/base?view=revision&revision=300421 as this code is only used on FreeBSD

stapelberg: Can you put this link into your commit message and mention that this change is to be consistent with the kernel, please?

bapt: done

PR #198, opened 2017-01-31
Include sys/select.h on OpenBSD

jasperla:
Fixes a compilation error:
```
/usr/include/net80211/ieee80211_ioctl.h:339: warning: implicit declaration of function 'howmany'
/usr/include/net80211/ieee80211_ioctl.h:339: error: 'NBBY' undeclared here (not in a function)
/usr/include/net80211/ieee80211_ioctl.h:339: error: variably modified 'nr_rxmcs' at file scope
```

### DIFF
--- a/src/print_wireless_info.c
+++ b/src/print_wireless_info.c
@@ -49,6 +49,7 @@
 #include <sys/ioctl.h>
 #include <sys/socket.h>
 #include <net/if.h>
+#include <sys/select.h>
 #include <sys/types.h>
 #include <netinet/in.h>
 #include <netinet/if_ether.h>


src/print_wireless_info.c
```diff
@@ -49,6 +49,7 @@
 #include <sys/ioctl.h>
 #include <sys/socket.h>
 #include <net/if.h>
+#include <sys/select.h>
```
eplanet: Why not wrapping this specific OpenBSD code with the ad hoc macro?
```cpp
#ifdef __OpenBSD__
#include <sys/select.h>
#endif
```

jasperla: It is guarded by `__OpenBSD__`, a few lines above ;-)

eplanet: My bad, I just looked at the diff… :+1:

jasperla:
Cheers, thanks guys!

PR #201, opened 2017-02-07
Suppress printing :00 seconds of remaining battery lifetime

jasperla:
As apm(4)'s estimate only has a granularity of minutes.

From tb@openbsd.org


Airblader:
Thanks!

PR #202, opened 2017-02-07
Plug an fd leak in the OpenBSD-specific code for _first_

jasperla:
Depending on user limits it would eventually run the user out of available file descriptors.


Airblader:
Thanks!

PR #204, opened 2017-02-19
Adding support for including the contents of a file in your status line.

enyalios:
This is useful to be able to have slow running processes update files that can be quickly included in your status line.   This can be used to display the weather in i3status.


Airblader:
#89, #19

Issue #205, opened 2017-02-24
Allow to print ipv6 in ethernet and wireless modules' format_up.


Dettorer:
Allowing something like this
`format_up = "eth0: %ip, %ipv6 (%speed)`

The code to get the ipv6 address could be similar to the ipv6 module's one.

Airblader:
What's wrong with using the ipv6 module?

Dettorer:
That module only shows the outgoing IPv6 address for global internet access. That's nice but it would be even better to be able to print the IPv6 address of other interfaces, which may be connected to different (local) networks.

I plan on trying to implement that feature myself, if it seems sensible to you.

stapelberg:
Network interfaces commonly have a plethora of different IPv6 addresses:
* a link-local address (which may be the only address, and hence useful)
* a site-local address (handed out by default by OpenWrt, for example)
* a global address
* a global address with privacy extension
* n previous global addresses which are kept around due to still being used in TCP connections

I have two concerns about this feature request:
1. Intuitively, I don’t see how we could select the appropriate address for the appropriate context.
2. Given the strong dependence on context, the resulting rather narrow usefulness makes the feature not pass the i3status philosophy rule of thumb: I wouldn’t look at the IP address(es) regularly anyway (like, say, the clock), especially since the global connectivity use-case is already handled by the ipv6 module.

Dettorer:
That makes sense, my implementation only shows the first IPv6 found for a given interface.

A more sensible feature request would be an "%iface" format option for the IPv6 module, allowing the bar to show the interface corresponding to the outgoing address.

stapelberg:
Showing the interface in the ipv6 module sounds good to me.

Dettorer:
Thank you for the feedback, I'm working on it.

eplanet:
@Dettorer have you got any update on that issue?

Dettorer:
Oh I completely forgot about that issue. Thank you for the ping, I should have the time to look at it in the next few weeks.

panmau:
Is there any update on this?

stapelberg:
> Is there any update on this?

I have asked in https://github.com/i3/i3status/pull/327#issuecomment-998081891 if the pull request author is still around and interested so that we can pick this back up.

Dettorer:
Implemented in #327.

PR #206, opened 2017-02-26
Allow to print IPv6 addresses in ethernet and wireless modules

Dettorer:
Mentioned in issue #205, this allows format like:

    format_up = "eth: %ip [%ip6] (%speed)"

This works for the ethernet and the wireless modules.


Airblader:
Putting aside the details of the change (style etc.), this looks OK to me and I'm convinced by the usecase. But given that we're particularly careful with i3status, I'd appreciate if @stapelberg could at least give a yes/no on the feature. Then I can do a detailed review.

stapelberg:
Commented on #205.

Issue #199, opened 2017-01-31
Feature Request: show_cpu support for multi-cpu/multicore systems


beeblebrox:
Thanks for the great work on the i3* line. 

Would it be possible to put multicore/cpu support on the roadmap for i3status? A configuration such as "cpu_usage all", "cpu_usage 0", cpu_usage 1", etc. would be nice, but I could see extending the formatter to accept something similar to %cpu<N>.

Airblader:
There's no such thing as a roadmap. We discuss whether a feature would be accepted and then wait for anybody to implement it. :-)

beeblebrox:
Makes sense, would such a feature be accepted?

eplanet:
That feature totally make sense to me, I'm even surprised nobody has ever come along with it until now… If you are not willing to implement it, I'd happily propose a PR.

Airblader:
I agree that it makes sense. But the decision is up to @stapelberg.

eplanet:
@Airblader: of course ;-)

beeblebrox:
@eplanet If you want to propose the PR don't let me get in your way, it is much appreciated. 

Unfortunately, even though this seems fairly straightforward I don't have the spare time needed to accomplish this; I would do it otherwise.

stapelberg:
The feature seems fine to me, pull requests welcome.

eplanet:
This issue can be closed now thanks to #209.

PR #209, opened 2017-03-08
Multiple CPU support for cpu_usage

eplanet:
This PR addresses the issue #199 asking for multiple CPU support. It takes an arbitrary CPU number and outputs its usage using the same arithmetic as for CPU aggregated value.

I works perfectly well, but please share your thoughts on the format option that is added: `%cpu<N>`, and its memory management of each CPU stat info that is definitely leaking. I couldn't figure out how to deal with it without big refactoring, for instance with argument passing to `print_cpu_usage` from main loop, but I think it's not the best option either. Last, since I don't have FreeBSD installed, I couldn't make it compatible for this platform, but at least the previous behaviour still works.

If you'd prefer another way, I'd be happy to modify it accordingly.


Airblader:
I haven't yet looked at the code. Just thoughts on the description:

> but please share your thoughts on the format option that is added

I think that's fine.

> and its memory management of each CPU stat info that is definitely leaking

This definitely isn't. We cannot merge a PR that knowingly introduced memory leaks. If it really requires larger changes to implement this cleanly (which also is a requirement) we should perhaps first have PRs to improve the situation and then tackle the feature?

> Last, since I don't have FreeBSD installed, I couldn't make it compatible for this platform, but at least the previous behaviour still works.

That's fine for now, but should be documented accordingly.

eplanet:
Thanks for your comments.

I actually think that the memory leak is acceptable since it is a static variable that gets dynamically allocated only once, a lot of program assume it only gets cleaned when the process exits anyway, so why bother. I would understand it's not 100% clean, so there are 2 obvious ways for deleting this memory leak:
- Allocating the memory in the main loop in `i3status.c`, which would mean that the state of `cpu_usage` is stored in the main loop instead of its module
- Allocating a knowingly large maximum number of CPU for static allocation, say 100, that would be pointless to render in i3status anyway

Please tell me which one is your favorite approach.

eplanet:
@stapelberg @Airblader could you give me a hint on the preferred approach? I think having a static pointer that is dynamically allocated totally acceptable, since it's freed on the process exit by kernel. It shouldn't be considered a memory leak per say.

stapelberg:
A dynamically allocated static pointer sounds reasonable to me.

eplanet:
Thanks @stapelberg. I think you can now consider merging this PR for it is working to me.

Airblader:
@stapelberg Yeah, I'd appreciate if you reviewed it. You're more knowledgable on this and lately I'm pretty busy (unfortunately).

src/print_cpu_usage.c
```diff
@@ -33,36 +33,92 @@
...
+static struct cpu_usage *prev_cpus = NULL;  // Shall leak on exit
+
+#ifndef PROC_LINE_LENGTH
+#define PROC_LINE_LENGTH 512
+#endif
+
+/*
+ * Gets the number of CPU from /proc/stat
+ */
+int get_cpu_count(const char *path) {
```
stapelberg: nit: make this a `static` function, it’s not used outside this file.

Also, another way to get the number of CPUs is the [sched_getaffinity(2)](https://manpages.debian.org/jessie/manpages-dev/sched_setaffinity.2) function. I think we should prefer that function as it uses fewer syscalls to achieve the same effect, which is a stated goal of i3status.

I further wonder if it’s possible to detect whether CPU hotplug is supported on the machine at all, and short-circuit calling this function at all.

src/print_cpu_usage.c
```diff
@@ -33,36 +33,92 @@
...
+    int cpu_count = -1;
+    FILE *file = fopen(path, "r");
+    if (file != NULL) {
```
stapelberg: to reduce the amount of indented code, use:

```c
if (file == NULL) {
  return 1;
}
// …
```

That might also allow you to get rid of the `cpu_count-1` quirk. If not, please explain what that’s for — i.e., why not just use a `while` loop (instead of `do … while`)?

src/print_cpu_usage.c
```diff
@@ -33,36 +33,92 @@
...
-void print_cpu_usage(yajl_gen json_gen, char *buffer, const char *format, const char *format_above_threshold, const char *format_above_degraded_threshold, const float max_threshold, const float degraded_threshold) {
+void print_cpu_usage(yajl_gen json_gen, char *buffer, const char *format, const char *format_above_threshold, const char *format_above_degraded_threshold, const char *path, const float max_threshold, const float degraded_threshold) {
     const char *selected_format = format;
     const char *walk;
     char *outwalk = buffer;
-    int curr_user = 0, curr_nice = 0, curr_system = 0, curr_idle = 0, curr_total;
+    struct cpu_usage curr_all = {0, 0, 0, 0, 0};
     int diff_idle, diff_total, diff_usage;
+    int cpu_count = get_cpu_count(path);
+    struct cpu_usage *curr_cpus = (struct cpu_usage *)malloc(cpu_count * sizeof(struct cpu_usage));
```
stapelberg: Could we move this dynamic allocation out of the hot path please?

src/print_cpu_usage.c
```diff
@@ -33,36 +33,92 @@
...
+            prev_cpus[i].total = 0;
+        }
+    }
+
 #if defined(LINUX)
     static char statpath[512];
-    char buf[1024];
-    strcpy(statpath, "/proc/stat");
+    char buf[4096];
+    strcpy(statpath, path);
```
stapelberg: I realize the `strcpy` was already in the code you’re touching, but we should be able to directly pass `path` to `slurp`, right?

src/print_cpu_usage.c
```diff
@@ -33,36 +33,92 @@
...
-        sscanf(buf, "cpu %d %d %d %d", &curr_user, &curr_nice, &curr_system, &curr_idle) != 4)
+        sscanf(buf, "cpu %d %d %d %d", &curr_all.user, &curr_all.nice, &curr_all.system, &curr_all.idle) != 4)
         goto error;
+    // Parsing all cpu values using strtok
+    if (strtok(buf, "\n")) {
+        int cpu_idx = 0;
+        char *buf_itr = strtok(NULL, "\n");
+        while (cpu_idx < cpu_count && buf_itr) {
+            if (sscanf(buf_itr, "cpu%d %d %d %d %d", &cpu_idx, &curr_cpus[cpu_idx].user, &curr_cpus[cpu_idx].nice, &curr_cpus[cpu_idx].system, &curr_cpus[cpu_idx].idle) != 5) {
+                fprintf(stderr, "Could not parse CPU number %d stat information\n", cpu_idx);
```
stapelberg: Use `goto error` for consistency with the existing error handling

src/print_cpu_usage.c
```diff
@@ -33,36 +33,92 @@
...
     if (!slurp(statpath, buf, sizeof(buf)) ||
-        sscanf(buf, "cpu %d %d %d %d", &curr_user, &curr_nice, &curr_system, &curr_idle) != 4)
+        sscanf(buf, "cpu %d %d %d %d", &curr_all.user, &curr_all.nice, &curr_all.system, &curr_all.idle) != 4)
         goto error;
+    // Parsing all cpu values using strtok
+    if (strtok(buf, "\n")) {
+        int cpu_idx = 0;
+        char *buf_itr = strtok(NULL, "\n");
+        while (cpu_idx < cpu_count && buf_itr) {
+            if (sscanf(buf_itr, "cpu%d %d %d %d %d", &cpu_idx, &curr_cpus[cpu_idx].user, &curr_cpus[cpu_idx].nice, &curr_cpus[cpu_idx].system, &curr_cpus[cpu_idx].idle) != 5) {
```
stapelberg: Could you refactor the code so as to not repeat the `sscanf` call please?

src/print_cpu_usage.c
eplanet: I was actually inspired by [htop method](https://github.com/hishamhm/htop/blob/master/linux/LinuxProcessList.c#L208) to get number of CPU. I do however agree that it might not be the best approach.

From what I read in the `sched_getaffinity(2)` doc you provided, I'm not sure its purpose is to return the actual number of available hardware CPU. I found an alternative that you might prefer, that reads `/sys/devices/system/cpu/online`: [sysconf(3)](https://linux.die.net/man/3/sysconf).

```cpp
int cpu_count = sysconf(_SC_NPROCESSORS_ONLN);
```

src/print_cpu_usage.c
eplanet: Using [sysconf(3)](https://linux.die.net/man/3/sysconf) would simplify that.

src/print_cpu_usage.c
eplanet: Just to be sure, this would mean we don't free the struct and allocate it with prev_cpus?

src/print_cpu_usage.c
eplanet: I can confirm it works without `strcpy`, meaning we can spare memory by deleting `statpath` declaration as well.

src/print_cpu_usage.c
eplanet: My bad!

src/print_cpu_usage.c
eplanet: If you mean by assuming line `n` matches cpu `n - 1`, totally.

eplanet:
Thank you very much for the review, I'm adjusting what we discussed, and I'll push the corrected version in a few minutes.

src/print_cpu_usage.c
```diff
@@ -33,36 +33,82 @@
...
-        sscanf(buf, "cpu %d %d %d %d", &curr_user, &curr_nice, &curr_system, &curr_idle) != 4)
+
+    // Detecting if CPU count has changed
+    int curr_cpu_count = sysconf(_SC_NPROCESSORS_ONLN);
+    if (curr_cpu_count != cpu_count) {
+        cpu_count = curr_cpu_count;
+        if (prev_cpus != NULL) {
+            free(prev_cpus);
+        }
+        prev_cpus = (struct cpu_usage *)malloc(cpu_count * sizeof(struct cpu_usage));
```
stapelberg: Use `calloc(cpu_count, sizeof(struct cpu_usage))` to initialize the memory with 0s, then get rid of the `for` loop below.

src/print_cpu_usage.c
stapelberg: curr_all should be equal to the sum, so can’t we just use `curr_all.user += curr_cpus[cpu_idx].user` etc.?

eplanet: I assume we could, but it would save only one sscanf syscall. Is it worth it?

stapelberg: Yes. It’s not so much about the extra syscall, but rather about reducing duplication in the code.

src/print_cpu_usage.c
```diff
@@ -33,36 +33,75 @@
...
-    char buf[1024];
-    strcpy(statpath, "/proc/stat");
-    if (!slurp(statpath, buf, sizeof(buf)) ||
-        sscanf(buf, "cpu %d %d %d %d", &curr_user, &curr_nice, &curr_system, &curr_idle) != 4)
+
+    // Detecting if CPU count has changed
+    int curr_cpu_count = sysconf(_SC_NPROCESSORS_ONLN);
+    if (curr_cpu_count != cpu_count) {
+        cpu_count = curr_cpu_count;
+        if (prev_cpus != NULL) {
```
stapelberg: Here and elsewhere, remove the checks: `free(x)` is specified as a no-op when `x == NULL`

eplanet: Indeed, good catch.

src/print_cpu_usage.c
eplanet: I learned something very interesting here :-).

src/print_cpu_usage.c
eplanet: That makes sense, I modified the code accordingly.

src/print_cpu_usage.c
```diff
@@ -33,36 +33,69 @@
...
+        prev_cpus = (struct cpu_usage *)calloc(cpu_count, sizeof(struct cpu_usage));
+        free(curr_cpus);
+        curr_cpus = (struct cpu_usage *)calloc(cpu_count, sizeof(struct cpu_usage));
+    }
+
+    char buf[4096];
+    if (!slurp(path, buf, sizeof(buf)))
         goto error;
+    // Parsing all cpu values using strtok
+    if (strtok(buf, "\n")) {
```
stapelberg: If this `strtok` call fails, we might as well `goto error`, right? No values will be presented, even if we continue running.

src/print_cpu_usage.c
```diff
@@ -33,36 +33,69 @@
...
+        int cpu_idx = 0;
+        char *buf_itr = strtok(NULL, "\n");
+        while (cpu_idx < cpu_count && buf_itr) {
```
stapelberg: You can pull this into a `for` loop to remove the duplicated `strtok` call:

```c
char *buf_itr = NULL;
for (int cpu_idx = 0; cpu_idx < cpu_count && (buf_itr = strtok(NULL, "\n")); cpu_idx++) {
// …
}
```

testcases/010-cpu-usage/expected_output.txt
```diff
@@ -0,0 +1 @@
+all: 02% CPU_0: 02% CPU_1: 02%
```
stapelberg: I think it’d be better if the usage numbers were different for the two CPUs so that the test covers more cases.

eplanet: I added some more test cases to better cover various situations

src/print_cpu_usage.c
eplanet: That totally makes sense!

src/print_cpu_usage.c
eplanet: Indeed. I fixed it.

stapelberg:
Looks good now. Could you clang-format it please?

eplanet:
Done! :relieved:. Thanks for the review!

PR #211, opened 2017-03-10
disk: don't remove trailing slash if path is a single character

### LINKED ISSUE #210: Root as disk path doesn't work
Commit 554db9f automatically removes a trailing slash from the path specified in the `disk` directive. However, it doesn't check if the path is just "/", so the following config is broken:

```
order += "disk /"

disk "/" {
    format = "%avail"
}
```

ixjlyons:
Fixes #210


Airblader:
This looks good to me, thank you. And sorry for the delay in review!

PR #213, opened 2017-03-22
Show IP address when address has a label

dlrobertson:
If an address has been added with a label, ensure the correct address is shown.

If a user has added a address with a description (e.g. `ip link addr add <addr>/<mask> dev <iface> label <iface>:<description>`) to the link "no IP" will be returned by `get_ip_addr`.

**Note:** Think of this more as an issue with a possible patch than an actual PR. Comments, critiques, and/or closing this PR for another fix are welcomed. I use `i3status` and hit this issue. There are definitely other (likely better) ways to solve this problem. For example, if the config could accept a label that would also solve the problem. AFAIK `get_ip_addr` would return the right address given `"<iface>:<label>"`

### DIFF
--- a/src/print_ip_addr.c
+++ b/src/print_ip_addr.c
@@ -29,6 +29,7 @@ const char *get_ip_addr(const char *interface, int family) {
 
     struct ifaddrs *ifaddr, *addrp;
     bool found = false;
+    int interface_len = strlen(interface);
 
     getifaddrs(&ifaddr);
 
@@ -39,13 +40,13 @@ const char *get_ip_addr(const char *interface, int family) {
     for (addrp = ifaddr;
 
          (addrp != NULL &&
-          (strcmp(addrp->ifa_name, interface) != 0 ||
+          (strncmp(addrp->ifa_name, interface, interface_len) != 0 ||
            addrp->ifa_addr == NULL ||
            addrp->ifa_addr->sa_family != family));
 
          addrp = addrp->ifa_next) {
         /* Check if the interface is down */
-        if (strcmp(addrp->ifa_name, interface) != 0)
+        if (strncmp(addrp->ifa_name, interface, interface_len) != 0)
             continue;
         found = true;
         if ((addrp->ifa_flags & IFF_RUNNING) == 0) {


eplanet:
Could you provide an example of output with and without your patch? I don't really understand what issue it is supposed to address, so it would help me a lot :-).

danlrobertson:
No problem!

## Minimal Example

Given the following config

```
ethernet test0 {
    format_up = "test0: %ip"
    format_down = "test0: down"
}
```

If `test0` were to have a labeled ip address

```
# ip link add name test0 type dummy
# ip link set dev test0 up
# ip link addr add 10.0.0.2/25 dev test0 label test0:dummyiface
```

The status bar pre patch would show `test0: no IP` and `test0: 10.0.0.2` post patch.

## Root Cause

The `ifaddrs` structure with the information we're interested in has `ifa_name == "test0:dummyiface"`. As a result, pre-patch [`strcmp(addrp->ifa_name = "test0:dummyiface", interface = "test0") == 0`](https://github.com/i3/i3status/pull/213/files#diff-f62c646a96e3079349936830133f6b3dL37) will never hold. The patch exploits the fact that we know that the address label will always be of the form `<iface>:<label>` due to some backwards compatibility issues.

## Alternative Solution

IMO the *right* way to do this is probably to allow something like

```
ethernet test0:dummyiface {
    ...
}
```

Or allow users to specify a label or interface and label combination in the config, but this patch is a viable workaround.

Let me know if this is not clear and I can grab some screenshots or something

Airblader:
I can't comment on the issue at hand, but let's not implement workarounds but proper solutions only.

danlrobertson:
> let's not implement workarounds but proper solutions only.

Sorry, workaround was probably a strong word. Depends on what your end goal is. If you want the user to be able to put in an interface and it "just works" this is it, but I'd prefer this to be in the config so that a user can choose the label to be shown.

Airblader:
@eplanet What are your thoughts?

eplanet:
Sorry about the delay for my answer... I tried your scenario and it works. I'm not very familiar with the `ip` command but the modification seems reasonable if it addresses your use case.

I think this is a good opportunity to add a test case, since the commands to create a label could be executed in the Docker image running the tests.

However I doubt it's very common, the decision to merge is up to @Airblader and @stapelberg anyway :smile:.

Airblader:
Since this has been laying around for a long time (sorry), I'll tag this with needs review + help wanted. I'm not very familiar with this, so some insights on the technical background would be appreciated.

dlrobertson:
The gist of it is that if a interface is assigned a static address with a label, the iface name associated with the address will be `interface:label` which means `strcmp(name from config, interface name)` won't work and therefore the address will not be shown.

Airblader:
Thanks! Is there a reason why we can't juse use `strcasecmp` with explicitly passing the length as the last argument? Then we wouldn't need this hand-stricked function, right?

Airblader:
Can you change the PR to that, please? I think there's no reason to not use that much simpler solution. :-)

dlrobertson:
Updated and ran basic tests to ensure that it works using `strncmp`

Airblader:
Thanks!

Issue #215, opened 2017-04-03
Can't parse axp288_fuel_gauge uevent


danielotero:
I have an Intel Atom table with a AXP288 PMIC that reports the battery information in the following way:
```
/sys/class/power_supply/axp288_fuel_gauge/uevent:

POWER_SUPPLY_NAME=axp288_fuel_gauge
POWER_SUPPLY_STATUS=Discharging
POWER_SUPPLY_PRESENT=1
POWER_SUPPLY_HEALTH=Good
POWER_SUPPLY_VOLTAGE_MAX_DESIGN=4200000
POWER_SUPPLY_VOLTAGE_NOW=3872000
POWER_SUPPLY_VOLTAGE_OCV=3942000
POWER_SUPPLY_CURRENT_NOW=-928000
POWER_SUPPLY_CAPACITY=75
POWER_SUPPLY_CAPACITY_ALERT_MIN=0
POWER_SUPPLY_TECHNOLOGY=Li-ion
POWER_SUPPLY_CHARGE_FULL=6404944
POWER_SUPPLY_CHARGE_NOW=4816448
```

This causes that batt_info (at [this](https://github.com/i3/i3status/blob/94651257cea4ff419f9cf2143d93193d5a5ccb96/src/print_battery_info.c#L179) point) to be:
```
{ full_design = -3,
  full_last = 25133000,
  remaining = 20505191,
  present_rate = 3327552,
  seconds_remaining = -1,
  percentage_remaining = -1,
  status = CS_DISCHARGING }
```
Which then causes the code to go through [this](https://github.com/i3/i3status/blob/94651257cea4ff419f9cf2143d93193d5a5ccb96/src/print_battery_info.c#L502) path, that causes a "No battery" result.

Can't this be done in a better way?

eplanet:
This is totally linked to #214. Just wait for the PR #216 to be merged, and `POWER_SUPPLY_CAPACITY` will be exploited to display correct percentage.

itdaniher:
Ran into this issue with an AXP803 on my pinebook using an old build of i3status in Ubuntu 16.04. Happy to say this fixes the issue for me!

eplanet:
Great! Thanks for sharing @itdaniher

eplanet:
@danielotero could you confirm that it fixes the issue for you and possibly close it?

danielotero:
Sorry about the delay, I almost forgot about this issue. 

[...]

Issue #214, opened 2017-03-22
i3status battery won't parse Apple Magic Trackpad battery uevent


edio:
@Output of `i3 --moreversion 2>&- || i3 --version`:
```
Binary i3 version: 4.13 (2016-11-08) © 2009 Michael Stapelberg and contributors
Running i3 version: 4.13 (2016-11-08) (pid 1994)o abort…)
Loaded i3 config: /home/edio/.config/i3/config (Last modified: Fri 24 Feb 2017 09:56:52 AM CET, 2251835 seconds ago)

The i3 binary you just called: /usr/bin/i3
The i3 binary you are running: i3
```

### What I did:

Added to `.config/i3status/config`

```
battery 1 {
    format = "Touchpad: %percentage"
    format_down = "No touchpad"
    path = "/sys/class/power_supply/hid-70:cd:60:f7:a0:5c-battery/uevent"
    low_threshold = 10
}
```

This should have displayed battery status of my Apple Magic Trackpad

```
$ cat /sys/class/power_supply/hid-70:cd:60:f7:a0:5c-battery/uevent
POWER_SUPPLY_NAME=hid-70:cd:60:f7:a0:5c-battery
POWER_SUPPLY_PRESENT=1
POWER_SUPPLY_ONLINE=1
POWER_SUPPLY_CAPACITY=100
POWER_SUPPLY_MODEL_NAME=applemagictrackpad
POWER_SUPPLY_STATUS=Discharging
POWER_SUPPLY_SCOPE=Device
```

### What I saw:

i3status displayed "No touchpad"

### What I expected instead:

i3 status should have displayed actual battery status of my touchpad

eplanet:
Thanks for reporting the issue. Could you provide your i3status version and strace execution as well?
```bash
i3status -v
strace -o /tmp/i3status.log i3status --run-once
```

I see two issues here: accessing the battery file and parsing its content.

For accessing the file, that obviously seems to exist on your computer, there's no reason why the reading function used in [src/general.c#L24](https://github.com/i3/i3status/blob/master/src/general.c#L24) would fail, since `cat` would do exactly the same. The `strace` output should greatly help understand what's going on.

[...]

PR #216, opened 2017-04-04
Add battery capacity parsing

eplanet:
This is a PR to correct #214 (and #215). It gives the priority to previous behavior, by forcing the remaining percentage to `POWER_SUPPLY_CAPACITY` only when no other information is available.


src/print_battery_info.c
```diff
@@ -138,9 +138,13 @@ static bool slurp_battery_info(struct battery_info *batt_info, yajl_gen json_gen
         if (BEGINS_WITH(last, "POWER_SUPPLY_ENERGY_NOW=")) {
             watt_as_unit = true;
             batt_info->remaining = atoi(walk + 1);
+            batt_info->percentage_remaining = -1;
         } else if (BEGINS_WITH(last, "POWER_SUPPLY_CHARGE_NOW=")) {
             watt_as_unit = false;
             batt_info->remaining = atoi(walk + 1);
+            batt_info->percentage_remaining = -1;
+        } else if (BEGINS_WITH(last, "POWER_SUPPLY_CAPACITY=") && batt_info->remaining == -1) {
```
Airblader: Doesn't this depend on the order in which the fields are in? Or is the order guaranteed?

eplanet: The order doesn't seem to be guaranteed, that's why I reset `batt_info->percentage_remaining` to `-1` in case either `POWER_SUPPLY_ENERGY_NOW` or `POWER_SUPPLY_CHARGE_NOW` are set. Doing that, it is guaranteed that `POWER_SUPPLY_CAPACITY` has the lowest priority.

But in the two cases reported by users, if `POWER_SUPPLY_CAPACITY` is set, the others are not.

Airblader: Ah, alright. I misread this the first time.

eplanet:
Great! Waiting for Michael's feedback then :smile:

stapelberg:
Thanks!

Issue #173, opened 2016-10-29
Ethernet port wrongly detected


MartijnBraam:
I have a laptop without an ethernet interface, only wireless. I use a usb ethernet interface with it. If the interface is not connected then the statusbar shows me that it exists but doesn't have an IP and has a speed of 10 Mbit/s. Sometimes it correctly recognizes that i don't have ethernet and show it as down. 

Everytime the interface is actually connected the status is fine.

``` bash
$ ip link
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
2: wlp2s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP mode DORMANT group default qlen 1000
    link/ether e8:2a:ea:76:64:34 brd ff:ff:ff:ff:ff:ff
3: sit0@NONE: <NOARP> mtu 1480 qdisc noop state DOWN mode DEFAULT group default qlen 1
    link/sit 0.0.0.0 brd 0.0.0.0
4: sixxs: <POINTOPOINT,MULTICAST,NOARP,UP,LOWER_UP> mtu 1280 qdisc fq_codel state UNKNOWN mode DEFAULT group default qlen 500
    link/none 
```

I have no idea what the next debugging step is. Do you need more information?

eplanet:
Could you both provide what `ls /sys/class/net` outputs on your laptop, and the paragraph of your `i3status.conf` related to network interface?

MartijnBraam:
The /sys/class/net output
```bash
ls -lah /sys/class/net
total 0
drwxr-xr-x  2 root root 0 29 okt 21:40 ./
drwxr-xr-x 52 root root 0 29 okt 09:44 ../
lrwxrwxrwx  1 root root 0 29 okt 09:44 lo -> ../../devices/virtual/net/lo/
lrwxrwxrwx  1 root root 0 29 okt 09:44 sit0 -> ../../devices/virtual/net/sit0/
lrwxrwxrwx  1 root root 0 29 okt 09:44 sixxs -> ../../devices/virtual/net/sixxs/
lrwxrwxrwx  1 root root 0 29 okt 09:44 wlp2s0 -> ../../devices/pci0000:00/0000:00:1c.3/0000:02:00.0/net/wlp2s0/
```

The config is just the defaults:


```
[...]

Issue #123, opened 2016-04-27
New naming convention of udev incompatible with `ethernet _first_`?


msis:
The newer version of `systemd`/`udev` uses new [naming schemes](https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/). Therefore, some times the Ethernet is named `enxXXXXXXXXXXXX` and is not detected by the the `ethernet _first_`.

stapelberg:
The code (https://github.com/i3/i3status/blob/dbc59d7376dd1c49b4948d17209910fad01b3560/src/first_network_device.c) does not make any assumptions about the device names. I’m not convinced that the problem at hand is the name of the device.

Can you describe which version of i3status you’re using, what you’re seeing and what you’re expecting?

msis:
So I'm using : 
```
i3status 2.10-19-gc094f47b4571 (2016-03-30) © 2008 Michael Stapelberg and contributors
```

When my ethernet connection is up, i.e. `ifconfig` returns an interface with the name `enxXXXXXXXXXXXX` with an IP address, `i3status` ignores it and returns `E: down`.

First I thought that it was the `path` variable limited to 1024. But I modified that on my fork and didn't seem to solve the problem. 

To make sure that interfaces is listed, I tried the example code provided through `man getifaddrs` and it lists it... and an `AF_INET`.

Airblader:
@msis It'd be easiest if you just debug the code referenced by @stapelberg to find out at which point the entry for your interface gets skipped. A simple way would be putting `fprintf` to each `continue`, or to use `gdb`.

msis:
Thanks guys!

After debugging, I have found out that virtual interface gets selected first before my real interfaces.

I googled around to see if there's any solution to skip virtual interfaces, and it seems that one solution would be to use the flags defined in `/usr/include/linux/if.h` and it's not good enough.
I don't know how useful it will be depending on the distrib users will be using. And also we need to rely on volatile flags, in my case `IFF_RUNNING` or `IFF_LOWER_UP` .

Do you guys have any idea how to avoid selecting a virtual interface?

stapelberg:
Which interface specifically gets selected in your case? I.e., what’s a virtual interface?

msis:
In the many case I tried it recently it's the docker local virtual network `docker0`.

stapelberg:
Maybe it would make sense to start a blacklist, or fill a whitelist with device names that udev assigns to physical ethernet interfaces?

[...]

PR #219, opened 2017-04-12
Add check for virtual ethernet devices

flammi:
The _first_ option for ethernet devices now uses the link in sysfs to determine
if it's a real device or just a virtual one (i.e veth** devices created by docker).

This might also fix the issues with tun/tap devices mentioned at the end of issue #123. Feedback is highly appreciated!

### DIFF
--- a/src/first_network_device.c
+++ b/src/first_network_device.c
@@ -57,8 +57,9 @@ static bool is_virtual(const char *ifname) {
 
     snprintf(path, sizeof(path), "/sys/class/net/%s", ifname);
     if ((target = realpath(path, NULL))) {
-        if (BEGINS_WITH(target, "/sys/devices/virtual/"))
+        if (BEGINS_WITH(target, "/sys/devices/virtual/")) {
             is_virtual = true;
+        }
     }
 
     free(target);
--- a/src/first_network_device.c
+++ b/src/first_network_device.c
@@ -50,6 +50,21 @@ static bool sysfs_devtype(char *dest, size_t n, const char *ifnam) {
     return true;
 }
 
+static bool is_virtual(const char *ifname) {
+    char path[1024];
+    char *target = NULL;
+    const char virtual_template[] = "/sys/devices/virtual/";
+
+    snprintf(path, sizeof(path), "/sys/class/net/%s", ifname);
+    if ((target = realpath(path, NULL))) {
+        if (strncmp(virtual_template, target, strlen(virtual_template)) == 0)
+            return true;
+    }
+    free(target);
+
+    return false;
+}
+
 static net_type_t iface_type(const char *ifname) {
 #ifdef __linux__
     char devtype[32];
@@ -57,6 +72,7 @@ static net_type_t iface_type(const char *ifname) {
     if (!sysfs_devtype(devtype, sizeof(devtype), ifname))
         return NET_TYPE_OTHER;
 
+    /* Default to Ethernet when no devtype is available */
     if (!devtype[0])
         return NET_TYPE_ETHERNET;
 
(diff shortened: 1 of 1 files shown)


src/first_network_device.c
```diff
@@ -50,13 +50,29 @@ static bool sysfs_devtype(char *dest, size_t n, const char *ifnam) {
     return true;
 }
 
+static bool is_virtual(const char *ifname) {
+    char path[1024];
+    char *target = NULL;
+    const char virtual_template[] = "/sys/devices/virtual/";
```
eplanet: This should be `static` to avoid allocation at call each.

src/first_network_device.c
```diff
@@ -50,13 +50,29 @@ static bool sysfs_devtype(char *dest, size_t n, const char *ifnam) {
...
+
+    snprintf(path, sizeof(path), "/sys/class/net/%s", ifname);
+    if ((target = realpath(path, NULL))) {
+        if (strncmp(virtual_template, target, strlen(virtual_template)) == 0)
```
eplanet: We usually use `BEGINS_WITH` macro for that purpose, it is available in that scope since defined in [i3status.h:32](https://github.com/i3/i3status/blob/master/include/i3status.h#L32).

src/first_network_device.c
```diff
@@ -50,13 +50,29 @@ static bool sysfs_devtype(char *dest, size_t n, const char *ifnam) {
...
+            return true;
```
eplanet: If you return before freeing `target`, you'll leak for each virtual path since `target` is guaranteed allocated in that scope.

src/first_network_device.c
flammi: Great! I wasn't aware that this macro exists, I included it in a new version and it makes the code a lot more readable.

src/first_network_device.c
flammi: Oh, thanks a lot. I completely overlooked that. Fixed in the next commit. Thanks a lot for looking at my code!

flammi:
Might also fix #173

src/first_network_device.c
```diff
@@ -50,13 +50,29 @@ static bool sysfs_devtype(char *dest, size_t n, const char *ifnam) {
...
 }
 
+static bool is_virtual(const char *ifname) {
+    char path[1024];
+    char *target = NULL;
+    bool is_virtual = false;
+
+    snprintf(path, sizeof(path), "/sys/class/net/%s", ifname);
+    if ((target = realpath(path, NULL))) {
+        if (BEGINS_WITH(target, "/sys/devices/virtual/"))
```
Airblader: Can you please use braces for the if statement?

Airblader:
@eplanet Are you happy with the PR now?

Airblader:
Please also leave a comment when you fixed that one last comment so I get notified. Thanks! :-)

src/first_network_device.c
flammi: Of course :-) I wasn't sure about your coding guidelines for this. The rest of the code of i3status skips the braces for if-statement with only one statement in the conditional block...

flammi:
Braces added! :-)

eplanet:
@Airblader I'm very happy indeed. Let's rebase and merge! Thanks @flammi for the PR.

Airblader:
Thanks!

eplanet:
What a pity, it has been merged without rebase.

Airblader:
Sorry, I must've overlooked that. Why is the rebate necessary if it merged without conflict?

eplanet:
No problem with the merge, but IMHO for the sake of clarity, an "added braces" commit message is certainly confusing the git log. I'd have preferred the commit to be unique to what it fixes by squashing the two later commits.

I don't think adding a commit message for each required change by a review is a good practice.

But that's not very important... :-)

Airblader:
Ah, you are talking about squashing the commits. Yes, you right, we should've done that before; we do usually want this to have a clean history. My bad for forgetting this. :-)

PR #227, opened 2017-06-07
Add support for more hardware

jackrosenthal:
This PR encompasses two, very minor, changes to how the `print_battery_info` function works.

1. Not all battery hardware has a `POWER_SUPPLY_ENERGY_NOW` field, and in the case that it is not there `POWER_SUPPLY_ENERGY_AVG` is equivalent and can be substituted.
2. If battery hardware does not have a designed full capacity and only the last full capacity can be used, the (potentially default) setting for `last_full_capacity` should be ignored and the last full capacity should be used anyway, as to prevent an inevitable "No battery" message.

These two changes were made to allow support for the iBook G4's battery hardware, but I'm certain it applies to more that just that laptop (eg. other PowerBooks). It should not break compatibility with any other hardware.

### DIFF
--- a/man/i3status.man
+++ b/man/i3status.man
@@ -323,8 +323,9 @@ estimated to be empty. If you want to use the last full capacity instead of the
 design capacity (when using the design capacity, it may happen that your
 battery is at 23% when fully charged because it’s old. In general, I want to
 see it this way, because it tells me how worn off my battery is.), just specify
-+last_full_capacity = true+. You can hide seconds in the remaining time and
-empty time estimations by setting +hide_seconds = true+.
++last_full_capacity = true+. Systems which do not measure a designed capacity
+will assume +last_full_capacity = true+. You can hide seconds in the remaining
+time and empty time estimations by setting +hide_seconds = true+.
 
 If you want the battery percentage to be shown without decimals, add
 +integer_battery_capacity = true+.
--- a/src/print_battery_info.c
+++ b/src/print_battery_info.c
@@ -135,8 +135,7 @@ static bool slurp_battery_info(struct battery_info *batt_info, yajl_gen json_gen
         if (*walk != '=')
             continue;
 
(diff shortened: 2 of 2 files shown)


eplanet:
Thanks for the PR! Could you check, just in case, that #216 doesn't fix your problem already?

jackrosenthal:
#216 isn't the key to this... the `POWER_SUPPLY_ENERGY_NOW` field is not available for PowerBook batteries.

eplanet:
Ok, since you didn't post any sample of the content of your file, I just couldn't know. Maybe you can enhance your PR with these things:
* A [test case](https://github.com/i3/i3status/tree/master/testcases), which should be very easy to add for battery parsing since there are already some
  * This would provide a sample of your file by the way
* A guarantee that `POWER_SUPPLY_ENERGY_AVG` is only used if necessary, ie if 
`POWER_SUPPLY_ENERGY_NOW` is not available
  * Maybe find the specs that tells why `POWER_SUPPLY_ENERGY_AVG` should be preferred and for what models (I assume it's not only iBook G4 batteries)
* Squash all commits

eplanet:
@jackrosenthal : are you working on this? Do you need any help?

jackrosenthal:
I got busy with work and forgot about this... I'll get to it later this week.

jackrosenthal:
Sorry for the long delay on this... I've had some issues with the hardware I was using to test: essentially, with this code the battery meter occasionally seems to jump from a "pretty safe percentage" (say, 80%) to 0%... not sure if it's this code or the machine I'm using.

Currently locating another machine and battery so I can double check this. I guess this is what I deserve using hardware from 2003! :)

Airblader:
@jackrosenthal Are you going to get back to this?

Airblader:
I'm going to close it for now until we hear back, but I'm happy to reopen if you continue on this!

Issue #231, opened 2017-06-28
Fixed-width %bitrate for wireless


devkev:
On my system the wifi bitrate often jumps between values like "585 MB/s" and "526.5 MB/s".  This is jarring because the change in width affects the alignment of all the other fields.  The cause is the way that `print_bitrate()` in `src/print_wireless_info.c` uses `%g` to print the bitrate.  However there is code in that function to choose an appropriate divisor, meaning that 3 digits (with no fractional part) should always be sufficient.  Therefore it would be more appropriate to use a format of `%3.0f` instead.

Airblader:
Seems to be in line with similar decisions we made, so SGTM. @stapelberg Your thoughts?

stapelberg:
SGTM

Issue #230, opened 2017-06-28
User-selectable zero/space padding on wireless fields


devkev:
#166/#167 changed the zero-padding on `%quality` (in wireless) to space-padding.  However, I prefer zero-padding, because with space-padding I find the extra horizontal space (between `%quality` and the entry to its left) to be distracting.  It would be nice if the user could choose between zero- and space-padding (for fields where that makes sense).

Airblader:
This goes a bit too much into the direction of visual preferences where no side is clearly better. We're not looking to accomodate every possible scenario with i3status, so perhaps switching to another tool might be a better decision here. I'm inclined to say no to this one. @stapelberg What do you think?

devkev:
I can say that I'm not about to switch away to another tool (after being happy with i3status for 4 years now) just because I happen to prefer `%03d` to `%3d` for a couple of fields.  That said, I am aware of the downsides of introducing these "%tag variants", as well as the downsides of allowing full customisation (eg. user-specified formats), or of not allowing any customisation at all.

stapelberg:
Yeah, patching this locally seems viable enough (a `sed(1)` invocation might do the trick) to justify not accepting this feature.

PR #232, opened 2017-06-28
Fixed-width wireless %bitrate, and 0-padded variants %quality0, %signal0, %noise0, %bitrate0

devkev:
Possible implementation for #230 and #231.

I previously had a (terrible) go at this (http://cr.i3wm.org/patch/415/); I notice now that `print_bitrate()` has been included directly, making things a lot easier.

Not sure how you feel about the idea of appending `0` onto `%bitrate` et al to get zero-padded versions, but it seemed to me to be the best compromise between super-clean-but-invasive and super-dirty-hack.

### DIFF
--- a/src/print_wireless_info.c
+++ b/src/print_wireless_info.c
@@ -464,10 +464,15 @@ error1:
     return 0;
 }
 
+static const char *format_3d_s = "%3d%s";
+static const char *format_03d_s = "%03d%s";
+
 void print_wireless_info(yajl_gen json_gen, char *buffer, const char *interface, const char *format_up, const char *format_down) {
     const char *walk;
     char *outwalk = buffer;
     wireless_info_t info;
+    unsigned int advance;
+    const char *format;
 
     INSTANCE(interface);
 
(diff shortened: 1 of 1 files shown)


Airblader:
I'd prefer to have two separate PRs for two separate features. But let's discuss the features first anyway.

src/print_wireless_info.c
```diff
@@ -114,7 +117,15 @@ static void print_bitrate(char *buffer, int buflen, int bitrate) {
         scale = 'k';
         divisor = kilo;
     }
-    snprintf(buffer, buflen, "%g %cb/s", rate / divisor, scale);
+    snprintf(buffer, buflen, format, rate / divisor, scale);
+}
+
+static void print_bitrate(char *buffer, int buflen, int bitrate) {
```
eplanet: Why not accepting `format_bitrate` as an argument to the function? Then you can switch between the 2 constants instead of implementing 2 functions.

Airblader:
@devkev Sorry that it has been so long, but there has also been an open question for quite a while :-) Are you going to continue on this?

Airblader:
Closing it for now. We can always reopen once there's feedback.

Issue #238, opened 2017-08-24
testcases/010-cpu-usage fails


kousu:
I can't get this test to work on my machine:

```
[kousu@phi i3status]$ ./travis/run-tests.pl 
Testing test case '006-cpu-usage-max-threshold-format'… OK
i3status: Cannot read CPU usage
Testing test case '010-cpu-usage'… Failed!
[kousu@phi i3status]$ 
```

```
[kousu@phi i3status]$ ./i3status --run-once -c testcases/010-cpu-usage/i3status.conf 
i3status: Cannot read CPU usage
cant read cpu usage
[kousu@phi i3status]$ 
```

Here's an `strace`:
```
[kousu@phi i3status]$ strace ./i3status --run-once -c testcases/010-cpu-usage/i3status.conf 
execve("./i3status", ["./i3status", "--run-once", "-c", "testcases/010-cpu-usage/i3status"...], 0x7ffc79706ce8 /* 47 vars */) = 0
brk(NULL)                               = 0x56243256b000
access("/etc/ld.so.preload", R_OK)      = -1 ENOENT (No such file or directory)
open("/etc/ld.so.cache", O_RDONLY|O_CLOEXEC) = 3
fstat(3, {st_mode=S_IFREG|0644, st_size=194477, ...}) = 0
mmap(NULL, 194477, PROT_READ, MAP_PRIVATE, 3, 0) = 0x7f9c19821000
close(3)                                = 0
open("/usr/lib/libconfuse.so.2", O_RDONLY|O_CLOEXEC) = 3
read(3, "\177ELF\2\1\1\0\0\0\0\0\0\0\0\0\3\0>\0\1\0\0\00000\0\0\0\0\0\0"..., 832) = 832
fstat(3, {st_mode=S_IFREG|0755, st_size=55040, ...}) = 0
mmap(NULL, 8192, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f9c1981f000
mmap(NULL, 2150816, PROT_READ|PROT_EXEC, MAP_PRIVATE|MAP_DENYWRITE, 3, 0) = 0x7f9c19420000
mprotect(0x7f9c1942c000, 2097152, PROT_NONE) = 0
mmap(0x7f9c1962c000, 8192, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED|MAP_DENYWRITE, 3, 0xc000) = 0x7f9c1962c000
[...]

Issue #218, opened 2017-04-09
CPU unit test fail


eplanet:
I noticed this problem while running unit tests on my laptop (4 CPU) instead of my usual i3status development virtual machine (2 CPU). Tests `010-cpu-usage` and `011-cpu-usage` fail if not run on a machine with exactly 2 CPU, since [sysconf(_SC_NPROCESSORS_ONLN)](https://github.com/i3/i3status/blob/master/src/print_cpu_usage.c#L65) will return something else than `2` and then fail the unit tests.

It shouldn't be urgent to fix it since it's specific to unit testing and luckily Travis runs on 2 CPU. But it might be a problem that pops in the future for contributors running tests or if Travis changes its default CPU count.

These two tests were introduced by #209.

PR #239, opened 2017-08-25
Fix CPU unit test case

eplanet:
This is the easiest fix I could think of for #218 (and #238). It adds a new setup feature for test cases that will generate the stat file according to the number of CPU available on the system.


testcases/010-cpu-usage/setup.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
Airblader: Please use 

    #!/usr/bin/env bash

here and below.

testcases/010-cpu-usage/setup.sh
```diff
@@ -0,0 +1,16 @@
...
+
+OUTPUT_FILE=$1/stat
```
Airblader: We should validate / sanitize the argument to avoid attempting to write in / if it isn't set.

Airblader:
@stapelberg What do you think about the approach to use a bash script?

stapelberg:
> @stapelberg What do you think about the approach to use a bash script?

We don’t currently use shell in i3status at all, so it would be good to implement this in Perl for consistency.

eplanet:
Thanks for your comments. I'll update the PR asap. I suppose we should convert [`expected_output.sh`](https://github.com/eplanet/i3status/blob/fix_cpu_tests/travis/run-tests.pl#L22-L23) to Perl as well? It's not used yet anyway.

stapelberg:
Yes please.

eplanet:
OK guys I updated the PR accordingly. Feel free to add any comment.

testcases/010-cpu-usage/setup.pl
```diff
@@ -0,0 +1,21 @@
...
+    say "Error with setup script: argument not provided or not a directory";
+    exit 1;
+}
+
+chomp(my $cpu_count = `grep -c -P '^processor\\s+:' /proc/cpuinfo`);
+my $output_file = "$ARGV[0]/stat";
+open(my $fh, '>', $output_file) or die "Could not open file '$output_file' $!";
+print $fh "cpu  0 0 0 0 0 0 0 0 0 0\n";
+print $fh "cpu0 100 0 0 0 0 0 0 0 0 0\n";
+print $fh "cpu1 50 0 0 50 0 0 0 0 0 0\n";
```
stapelberg: Are you assuming that all systems have at least two logical cores here? It seems more prudent to only assume one core is present.

travis/run-tests.pl
```diff
@@ -8,14 +8,19 @@
 
 sub TestCase {
     my ($dir) = @_;
+
+    if ( -f "@_/setup.pl") {
+        system($^X, "@_/setup.pl", ($dir));
```
stapelberg: Here and below, use the more descriptive `$EXECUTABLE_NAME` instead of `$^X` :). This might require adding `use English;` at the top.

travis/run-tests.pl
```diff
@@ -8,14 +8,19 @@
...
     my $conf = "$dir/i3status.conf";
     my $testres = `./i3status --run-once -c $conf`;
     my $refres = "";
 
     if ( -f "@_/expected_output.txt") {
         $refres = `cat "@_/expected_output.txt"`;
-    } elsif ( -f "@_/expected_output.sh") {
-        $refres = `bash @_/expected_output.sh`;
+    } elsif ( -f "@_/expected_output.pl") {
+        $refres = capture($^X, "@_/expected_output.pl");
```
stapelberg: What is `capture`? Where is it defined? My `perldoc(1)` doesn’t know about it, and grepping the i3status repo yields no results.

testcases/010-cpu-usage/setup.pl
eplanet: Those tests were designed for #209 so it didn't make sense to support one core. But now I think we got all we need to support one core as well, I'll update that very soon.

travis/run-tests.pl
eplanet: You're right, that was an avoidable mistake 😖. It's defined in [Capture/Tiny.pm](http://search.cpan.org/~dagolden/Capture-Tiny-0.46/lib/Capture/Tiny.pm#capture), but using backticks seems to be better anyway.

eplanet:
I updated the PR to avoid assuming at least 2 cores, and added a `cleanup.pl` file in test cases to remove generated `stat` files. It must be good to go now 😄.

testcases/010-cpu-usage/cleanup.pl
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env perl
+
+use v5.10;
+use strict;
+use warnings;
+
+if ($#ARGV != 0 || not -d $ARGV[0]) {
+    say "Error with setup script: argument not provided or not a directory";
```
stapelberg: s/setup/cleanup/

testcases/010-cpu-usage/cleanup.pl
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env perl
+
+use v5.10;
+use strict;
+use warnings;
+
+if ($#ARGV != 0 || not -d $ARGV[0]) {
```
stapelberg: Here and elsewhere: more idiomatic to use `! -d $ARGV[0]` instead of `not -d $ARGV[0]` (the latter differs in precedence from what one would expect, IIRC)

testcases/010-cpu-usage/cleanup.pl
```diff
@@ -0,0 +1,15 @@
...
+    say "Error with setup script: argument not provided or not a directory";
+    exit 1;
+}
+
+my $output_file = "$ARGV[0]/stat";
+if ( -f $output_file) {
+    unlink $output_file;
+}
```
stapelberg: Here and elsewhere: please add a newline to the end of the file

testcases/010-cpu-usage/cleanup.pl
```diff
@@ -0,0 +1,15 @@
...
+use strict;
+use warnings;
+
+if ($#ARGV != 0 || not -d $ARGV[0]) {
+    say "Error with setup script: argument not provided or not a directory";
+    exit 1;
+}
+
+my $output_file = "$ARGV[0]/stat";
+if ( -f $output_file) {
```
stapelberg: nit: `if (-f $output_file)` (remove the whitespace)

travis/run-tests.pl
```diff
@@ -3,19 +3,29 @@
...
     my $conf = "$dir/i3status.conf";
     my $testres = `./i3status --run-once -c $conf`;
     my $refres = "";
 
     if ( -f "@_/expected_output.txt") {
         $refres = `cat "@_/expected_output.txt"`;
-    } elsif ( -f "@_/expected_output.sh") {
-        $refres = `bash @_/expected_output.sh`;
+    } elsif ( -f "@_/expected_output.pl") {
+        $refres = `/usr/bin/env perl @_/expected_output.pl`;
```
stapelberg: Why not `$EXECUTABLE_NAME @_/expected_output.pl` for consistency with the `system()` calls above and below?

eplanet: This is to catch `stdout` in `$refres` variable. With `system` what you catch is the return code.

stapelberg: Yes. What I was alluding to is your usage of `/usr/bin/env perl` instead of `$EXECUTABLE_NAME`.

eplanet:
I took your comments into account. I made my usual typos / mistakes... Sorry about that! I hope this time it's cool. Tested successfully on both multiple cores and solo core.

stapelberg:
Thanks!

kousu:
This works for me! Thanks for fixing it!

I didn't realize this would be such a convoluted fix. I thought it was just a missing newline. Thanks for tackling it so quickly :)

```
[kousu@phi i3status]$ ./travis/run-tests.pl 
Testing test case '006-cpu-usage-max-threshold-format'… OK
Testing test case '010-cpu-usage'… OK
Testing test case '015-battery-capacity'… OK
Testing test case '008-load-max-threshold-format'… OK
Testing test case '007-cpu-usage-degraded-threshold-format'… OK
i3status: Cannot read CPU usage
Testing test case '012-cpu-usage-error'… OK
Testing test case '011-cpu-usage'… OK
Testing test case '014-battery-capacity'… OK
Testing test case '003-path-exists'… OK
Testing test case '005-cpu-temp-max-threshold-format'… OK
i3status: Cannot read CPU usage
Testing test case '013-cpu-usage-error'… OK
Testing test case '002-cpu-temp'… OK
Testing test case '004-run-watch'… OK
Testing test case '001-battery'… OK
Testing test case '009-disk-low-threshold-format'… OK
[kousu@phi i3status]$ 
[kousu@phi i3status]$ git log --decorate
commit 884e4da6730a31fc7a5ef892fb9cb061baaf5f6c (HEAD -> master, origin/master, origin/HEAD)
Author: Emeric Planet <emeric.planet@datadoghq.com>
Date:   Tue Aug 29 19:01:30 2017 +0200

    Fix CPU unit tests (#239)
    
    Support any amount of available cores on testing machine.
```

eplanet:
Great news! Thanks for your feedback. It was indeed not so easy to end up with a nice fix, but that's the price to pay since the code is tightly coupled with system calls... :-P

Issue #242, opened 2017-09-03
Add FreeBSD first_eth_interface support & fix compilation


ericdahl:
While investigating why 2.11 did not resolve `ethernet _first_`, I found #197 which added OpenBSD support, but now results in a compilation failure on FreeBSD because there's [no implementation defined](https://github.com/i3/i3status/blob/6a19709/src/first_network_device.c#L130).

```bash
$ uname 
FreeBSD

$ gmake
src/first_network_device.c:130:2: error: Missing implementation to determine interface
      type.
#error Missing implementation to determine interface type.
 ^
1 error generated.
gmake: *** [Makefile:88: src/first_network_device.o] Error 1
```

I did not see any issue or PR tracking this.

Airblader:
Yes, the `#error` is not good. We should've preserved the old behavior for BSDs in the case of OpenBSD.

CC @jasperla

jasperla:
That `#error` should indeed be removed, please see #243

eplanet:
This issue should probably be closed since it's been fixed by jasperla.

Airblader:
Yes, thanks!

PR #243, opened 2017-09-04
Remove overly aggressive #error

jasperla:
This removes an `#error` that breaks FreeBSD as noted in #242

### DIFF
--- a/src/first_network_device.c
+++ b/src/first_network_device.c
@@ -82,8 +82,6 @@ static net_type_t iface_type(const char *ifname) {
 
     if (strcmp(devtype, "wwan") == 0)
         return NET_TYPE_OTHER;
-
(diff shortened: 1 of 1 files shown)


src/first_network_device.c
```diff
@@ -126,8 +126,6 @@ static net_type_t iface_type(const char *ifname) {
         close(s);
         return NET_TYPE_ETHERNET;
     }
-#else
-#error Missing implementation to determine interface type.
```
Airblader: Now this function won't return anything anymore, right? We should probably restore the old behavior of returning `NET_TYPE_OTHER`.

Airblader:
Sorry for the long wait, but as always thanks for the patch! Merging this so it's included in the upcoming release.

PR #245, opened 2017-09-07
Add module "hostname"

maride:
This PR just contains a small additional module, printing the hostname of the current machine.

This may seem unnecessary, however I found this module helpful while working with multiple VMs (with the same setup, thus looking exactly the same...).


Airblader:
Thanks for the PR and sorry for the long wait. I'm going to close this because my immediate thought is that this isn't in line with what we typically want as modules in i3status. If you do want to make a case for it, let's please discuss it in an issue first. Thanks!

PR #246, opened 2017-10-25
Add network test cases

eplanet:
These test cases might be useful to have for non regression.


eplanet:
@stapelberg : could you considere merging that harmless one? 😄

stapelberg:
Sure! Thank you.

Issue #240, opened 2017-08-27
Stop discriminating IPv6


telmich:
When running i3status and being in an IPv6 only network, i3status shows:

W: ( ... ) no IP

Having

order += "wireless wlp2s0"
order += "ipv6"

in my i3status.conf, I do see the IPv6 address in the next field. However seeing "no IP" next to it, simply does not make sense in the age of IPv6. There is an IP. A better than IPv4 IP address even!

I am aware that this is a bit like "Use Windows NT or better - so I used Linux" bug report, however I think it is time to fix the sentiment at the base of our hearts. IPv6 is here, IPv6 is not a second, but a first class citizen.

I suggest to

a) rephrase "no IP" to "no IPv4"
b) include IPv6 into the wireless block
c) OR exclude IPv4 from the wireless block

I think c) is not that feasible (breaking existing user), however I think b) is feasible if we exclude the link local address.

Btw, thanks for the simple status bar - it's really nice to use!

Output of i3 --moreversion 2>&- || i3 --version:

Binary i3 version: 4.13 (2016-11-08) © 2009 Michael Stapelberg and contributors
Running i3 version: 4.13 (2016-11-08) (pid 24983) abort…)
Loaded i3 config: /home/nico/.i3/config (Last modified: Thu 03 Aug 2017 06:01:08 PM CEST, 2047424 seconds ago)

The i3 binary you just called: /usr/bin/i3
The i3 binary you are running: i3

eplanet:
Well I'd say that IPv4 is far from dead and will endure for quite a long time, if not forever. I even personally prefer it, since it's less verbose and a lot easier to memorize / input.

That being said, I agree that when no IPv4 is available but an IPv6 is, it should be displayed (you "solution b"). Do you intend to propose a PR for that?

telmich:
After knowing which direction to go, I will have a look at the code and submit a PR soon!

eplanet:
Will wait for @stapelberg and @Airblader insights then.

stapelberg:
Note that the ipv6 module does something different: it displays whether you have an IPv6 address with global connectivity (by creating a connectionless UDP socket and obtaining the local address).

Displaying the most important IPv6 address when an interface doesn’t have an IPv4 address sounds good to me.

I think we should implement that with a new format string though, and then change the default config, so that existing users aren’t surprised by the new behavior.

eplanet:
@telmich #247 has been merged, could you try with latest master and confirms it works in your environment? Thanks!

[...]

PR #247, opened 2017-10-25
Add IPv6 address when IPv4 isn't available

eplanet:
This is an attempt to address #240. I didn't add a new format string, because however surprising it may be for existing users, I don't feel it would be unwanted. That's for you to decide for sure!

### DIFF
--- a/include/i3status.h
+++ b/include/i3status.h
@@ -215,7 +215,7 @@ void print_disk_info(yajl_gen json_gen, char *buffer, const char *path, const ch
 void print_battery_info(yajl_gen json_gen, char *buffer, int number, const char *path, const char *format, const char *format_down, const char *status_chr, const char *status_bat, const char *status_unk, const char *status_full, int low_threshold, char *threshold_type, bool last_full_capacity, bool integer_battery_capacity, bool hide_seconds);
 void print_time(yajl_gen json_gen, char *buffer, const char *title, const char *format, const char *tz, const char *locale, const char *format_time, time_t t);
 void print_ddate(yajl_gen json_gen, char *buffer, const char *format, time_t t);
-const char *get_ip_addr(const char *interface);
+const char *get_ip_addr(const char *interface, int family);
 void print_wireless_info(yajl_gen json_gen, char *buffer, const char *interface, const char *format_up, const char *format_down);
 void print_run_watch(yajl_gen json_gen, char *buffer, const char *title, const char *pidfile, const char *format, const char *format_down);
 void print_path_exists(yajl_gen json_gen, char *buffer, const char *title, const char *path, const char *format, const char *format_down);
--- a/man/i3status.man
+++ b/man/i3status.man
@@ -305,8 +305,9 @@ network interface found on the system (excluding devices starting with "lo").
 === Ethernet
 
 Gets the IP address and (if possible) the link speed of the given ethernet
-interface. Getting the link speed requires the cap_net_admin capability. Set
-it using +setcap cap_net_admin=ep $(which i3status)+.
+interface. If no IPv4 address is available and an IPv6 address is, it will be
+displayed. Getting the link speed requires the cap_net_admin capability.
+Set it using +setcap cap_net_admin=ep $(which i3status)+.
 
 The special interface name `_first_` will be replaced by the first non-wireless
 network interface found on the system (excluding devices starting with "lo").
--- a/src/print_eth_info.c
+++ b/src/print_eth_info.c
@@ -2,6 +2,7 @@
 #include <string.h>
 #include <limits.h>
 #include <stdio.h>
+#include <stdlib.h>
 #include <sys/ioctl.h>
 #include <sys/types.h>
 #include <sys/socket.h>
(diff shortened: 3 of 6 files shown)


eplanet:
@stapelberg : I'm totally open to comments on that PR, but it works fine on my side. Could you consider it when you have time? Thanks!

src/print_eth_info.c
```diff
@@ -121,22 +121,47 @@ static int print_eth_speed(char *outwalk, const char *interface) {
...
     INSTANCE(interface);
 
-    if (ip_address == NULL) {
-        START_COLOR("color_bad");
-        outwalk += sprintf(outwalk, "%s", format_down);
-        goto out;
+    static char ipv4_address[512], ipv6_address[512];
+    memset(ipv4_address, 0, sizeof(ipv4_address));
+    memset(ipv6_address, 0, sizeof(ipv6_address));
+    const char *ipv4_address_ptr = get_ip_addr(interface, AF_INET);
```
stapelberg: Instead of manually allocating, clearing and copying buffers, please use:
```c
char *ipv4_address = sstrdup(get_ip_addr(interface, AF_INET));
// …
free(ipv4_address);
```

src/print_eth_info.c
```diff
@@ -121,22 +121,47 @@ static int print_eth_speed(char *outwalk, const char *interface) {
...
+    if (ipv4_address_ptr != NULL)
+        memcpy(ipv4_address, ipv4_address_ptr, 512);
+    const char *ipv6_address_ptr = get_ip_addr(interface, AF_INET6);
+    if (ipv6_address_ptr != NULL) {
+        memcpy(ipv6_address, ipv6_address_ptr, 512);
+        char *prct_ptr = strstr(ipv6_address, "%");
```
stapelberg: Please add a comment clarifying the idea behind this code, which I assume is to strip interface identifiers as they’re already part of the i3status output.

src/print_eth_info.c
```diff
@@ -121,22 +121,47 @@ static int print_eth_speed(char *outwalk, const char *interface) {
...
+        if (prct_ptr != NULL)
+            *prct_ptr = '\0';
     }
 
-    if (BEGINS_WITH(ip_address, "no IP"))
-        START_COLOR("color_degraded");
-    else
+    bool prefer_ipv4 = true;
```
stapelberg: Instead of the helper, how about:
```c
const char *ip_address = (ipv4_address ? ipv4_address : ipv6_address);
if (ip_address == NULL) {
  START_COLOR("color_bad");
  outwalk += sprintf(outwalk, "%s", format_down);
  goto out;
}
```

eplanet: That's because the IP can contain "no IP", in which case we cannot avoid to test if IPv6 has a valid value.

The chosing between IPv4 and IPv6 looks like that matrix:

| ipv4_address | ipv6_address | Chosen IP | Color |
| ------------------|------------------|---------------|---------|
| NULL | NULL | None | bad (red) |
| NULL | no IP | IPv6 | degraded (orange) |
| NULL | ::1/128 | IPv6 | ok (green) |
| no IP | NULL | IPv4 | degraded |
| no IP | no IP | IPv4 | degraded |
| no IP | ::1/128 | IPv6 | ok |
| 127.0.0.1 | NULL | IPv4 | ok |
| 127.0.0.1 | no IP | IPv4 | ok |
| 127.0.0.1 | ::1/128 | IPv4 | ok |

I cannot think of a more elegant way to test all these conditions.

stapelberg: I see. Could you add this table (as a comment) to the code please?

eplanet:
Ok I think I got it. I did a little hack on local code to test the changes (see below). I think I'm going to work on another PR to provide unit testing on the code itself, rather than the testcases that are a sort of integration tests.

Feel free to comment further! And thanks for the advice.

```cpp
void print_eth_info(yajl_gen json_gen, char *buffer, const char *interface, const char *format_up, const char *format_down) {
    print_eth_info_r(json_gen, buffer, interface, format_up, format_down, NULL, NULL);
    print_eth_info_r(json_gen, buffer, interface, format_up, format_down, NULL, "no IP");
    print_eth_info_r(json_gen, buffer, interface, format_up, format_down, NULL, "::1/128%bla");
    print_eth_info_r(json_gen, buffer, interface, format_up, format_down, "no IP", NULL);
    print_eth_info_r(json_gen, buffer, interface, format_up, format_down, "no IP", "no IP");
    print_eth_info_r(json_gen, buffer, interface, format_up, format_down, "no IP", "::1/128%bla");
    print_eth_info_r(json_gen, buffer, interface, format_up, format_down, "127.0.0.1", NULL);
    print_eth_info_r(json_gen, buffer, interface, format_up, format_down, "127.0.0.1", "no IP");
    print_eth_info_r(json_gen, buffer, interface, format_up, format_down, "127.0.0.1", "::1/128%bla");
}
```

eplanet:
@stapelberg: could you consider the updated PR?

include/i3status.h
```diff
@@ -215,7 +215,7 @@ void print_disk_info(yajl_gen json_gen, char *buffer, const char *path, const ch
 void print_battery_info(yajl_gen json_gen, char *buffer, int number, const char *path, const char *format, const char *format_down, const char *status_chr, const char *status_bat, const char *status_unk, const char *status_full, int low_threshold, char *threshold_type, bool last_full_capacity, bool integer_battery_capacity, bool hide_seconds);
 void print_time(yajl_gen json_gen, char *buffer, const char *title, const char *format, const char *tz, const char *locale, const char *format_time, time_t t);
 void print_ddate(yajl_gen json_gen, char *buffer, const char *format, time_t t);
-const char *get_ip_addr(const char *interface);
+const char *get_ip_addr(const char *interface, unsigned short family);
```
stapelberg: Instead of “unsigned short”, can you just use “int” here please? Using the native integer size is usually preferable.

man/i3status.man
```diff
@@ -305,8 +305,9 @@ network interface found on the system (excluding devices starting with "lo").
 === Ethernet
 
 Gets the IP address and (if possible) the link speed of the given ethernet
-interface. Getting the link speed requires the cap_net_admin capability. Set
-it using +setcap cap_net_admin=ep $(which i3status)+.
+interface. If no IPv4 address is available and an IPv6 is, it will be displayed.
```
stapelberg: nit: s/IPv6/IPv6 address/ please

src/print_eth_info.c
```diff
@@ -118,25 +119,57 @@ static int print_eth_speed(char *outwalk, const char *interface) {
...
     INSTANCE(interface);
 
-    if (ip_address == NULL) {
-        START_COLOR("color_bad");
-        outwalk += sprintf(outwalk, "%s", format_down);
-        goto out;
+    char *ipv4_address = sstrdup(get_ip_addr(interface, AF_INET));
+    char *ipv6_address = sstrdup(get_ip_addr(interface, AF_INET6));
+
+    // Removing '%' and following characters from IPv6
```
stapelberg: This comment documents the “what”, not the “why”. Can you append e.g. “: the interface identifier is redundant, as the output already includes the interface name”

nit: use `/* */`, not `//`

include/i3status.h
eplanet: I was misled by the [`sockaddr`](https://www.gnu.org/software/libc/manual/html_node/Address-Formats.html) struct which actually was `short int` anyway... Corrected!

eplanet:
@stapelberg : thanks for your comments, that should be good to go now.

stapelberg:
Thanks!

PR #253, opened 2017-12-09
Call tzset at each time print

### LINKED ISSUE #223: Time doesn't update with system time
I use a dual boot system and sometimes have to update the time over the network. i3status doesn't update the time after this fact. Is there something that I can do to get i3status to update instead of restarting i3wm.

eplanet:
That syscall must be made at each time print since the timezone can be changed outside `i3status` and wouldn't be taken into account.

Fixes #223 and probably some other use cases.

### DIFF
--- a/src/print_time.c
+++ b/src/print_time.c
@@ -29,9 +29,9 @@ void set_timezone(const char *tz) {
         } else {
             unsetenv("TZ");
         }
-        tzset();
         current_timezone = tz;
     }
+    tzset();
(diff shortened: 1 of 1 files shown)


Airblader:
Thanks! I'll wait for @domenkozar's confirmation in #223 before merging.

domenkozar:
confirmed!

Airblader:
Thanks!

PR #254, opened 2017-12-11
Update comment as in #247

eplanet:
@stapelberg A thousand excuses... I thought I added that comment modification as requested in #247, but after double-checking I didn't! My bad!


stapelberg:
No worries, and thanks for sending the fix!

Issue #248, opened 2017-11-21
README SLOC Discrepancy


ajmeese7:
On the [i3 website](https://i3wm.org/i3status/) it says 'about 3000 SLOC' and in the [README here](https://github.com/i3/i3status/blob/master/README.md) it says 'about 1500 SLOC'. Which is correct?

eplanet:
I suppose the first one...

```
$ find -name "*.[c|h]" | xargs cat | sort | uniq | wc -l
3066
```

eplanet:
You can close that issue I suppose. Thanks!

Airblader:
Yes. Thanks. :-)

PR #255, opened 2017-12-11
Get rid of the SLOC comment

eplanet:
This is related to #248. Is it worth mentioning that fact anyway?


Airblader:
> Is it worth mentioning that fact anyway?

IMHO, no. Thanks!

PR #256, opened 2017-12-13
Read multiple batteries on DragonFly BSD.

oyvinht:
Using ACPI to read individual battery information instead of just the
accumulations that sysctl provide.

Like for OpenBSD, it requires the system dependency sys/fcntl.h.

No other dependencies added or changed.

### DIFF
--- a/src/print_battery_info.c
+++ b/src/print_battery_info.c
@@ -1,9 +1,9 @@
 // vim:ts=4:sw=4:expandtab
 #include <ctype.h>
-#include <time.h>
-#include <string.h>
-#include <stdlib.h>
 #include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
 #include <yajl/yajl_gen.h>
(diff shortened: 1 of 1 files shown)


Airblader:
I can't speak for whether it is correct, but since it affects only DragonFly, I'll rely on your expertise. However, could you please squash the commits? Thanks and sorry for the wait!

oyvinht:
Sure, no problem. I have squashed together all commits since fork now.

I can't make any guarantees for stability - as I have only tested it myself - but it has been fault free here on a regularly updated system (DragonFlyBSD master branch) for the last three-four months.

Airblader:
Thanks! I'll merge this after the i3status release. I do think it'd be good to have it run through some testing rather than instantly releasing it. :-)

oyvinht:
Yes, it (as always) definitively is.

Airblader:
I lost track of this a bit, I apologize. But let's go with it now. Thanks for the PR!

PR #259, opened 2018-01-28
Treat zero battery capacity as "not available"

johcgt:
`print_battery_info` computes `batt_info.percentage_remaining` by
dividing batt_info.remaining by `full`. If `full` is `0` then the
battery remaining will be reported as "inf".

Before this, it tries to set `full` to either the design capacity or to
the last known good charge. It determines if these values are available
by checking whether their fields in `batt_info` are non-negative. As it
initialized `batt_info` with values of `-1`, a non-negative value
implies that something has provided a value.

`slurp_all_batteries` and `add_battery_info` however initialize these
fields to zero, so if these functions are called then
`batt_info.full_design` will always be used.

This means that on systems that don't provide a value for design
capacity the percentage remaining will be reported as "inf", unless the
user has set `last_full_capacity` to `true` in their `i3status.conf`.

This patch changes `print_battery_info` to expect values for the battery
capacity to be strictly greater than zero. This seems reasonable as a
battery with a capacity of zero isn't useful.

An alternative solution would be to change `slurp_all_batteries` and
`add_battery_info` to initialize `batt_info` with `-1`, as
`print_battery_info` does. This is less appealing as `add_battery_info`
is accumulating the values, so using `-1` would introduce off-by-one
errors without additional code to avoid them.


Airblader:
Thanks for the detailed explanation. This LGTM, what does @stapelberg think?

PR #262, opened 2018-02-06
Add check for interval value

foutaise:
This PR adds a check for the interval value in the general section of the configuration.
A zero value generates a core dump on FreeBSD systems (division by zero):

```
$ cat i3status.conf
...
general {
        interval = 0
}
...

$ i3status -c i3status.conf
i3status: trying to auto-detect output_format setting
i3status: auto-detected "term"
♪: 80% | 0.12 | 2018-02-06 12:39:35
Floating point exception (core dumped)
```

With the patch applied:

```
$ i3status -c i3status.conf
i3status: trying to auto-detect output_format setting
i3status: auto-detected "term"
Invalid interval attribute found in section general, line 15: 0
Expected positive integer
```
Best Regards.

### DIFF
--- a/i3status.c
+++ b/i3status.c
@@ -631,6 +631,11 @@ int main(int argc, char *argv[]) {
         die("Could not create socket\n");
 
     int interval = cfg_getint(cfg_general, "interval");
+    if (interval <= 0) {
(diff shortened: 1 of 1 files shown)


Airblader:
Thanks!

PR #265, opened 2018-02-11
Print wireless quality without leading whitespace

tylerbrazier:
A config like
```
wireless wlan0 {
  format_up = "%ip (%quality)"
}
```
Will now result in something like `10.0.0.30 (85%)`
rather than `10.0.0.30 ( 85%)`

Also made this change for "signal" and "noise".


Airblader:
Thanks for the PR and sorry for the wait! The padding is intentional to avoid a flickering lengths in the statusline. I get that this is a matter of taste, but changing this would probably upset as many people as would be happy.

PR #266, opened 2018-02-12
OpenBSD: fix volume display on some hardware

moreais:
On some hardware the AUDIO_MIXER_READ ioctl requires the current number
of channels to be set, and the mute device found needs to be checked if
it belongs to the master output device.

This patch was committed to OpenBSD's ports tree: https://cvsweb.openbsd.org/cgi-bin/cvsweb/ports/x11/i3status/patches/patch-src_print_volume_c?rev=1.9&content-type=text/x-cvsweb-markup


Airblader:
Thanks!

PR #267, opened 2018-02-21
Add content like external IP address

TheMooseyOne:
Short script to add content and preserve color JSON formatting when using i3bar


Airblader:
Thanks for the PR! I don't think we're really looking to add to the contrib folder right now and it's also not very flexible anyway as it would have to be maintained here as well. Please publish this separately in a dedicated repository (if you want). Thanks!

PR #268, opened 2018-02-21
Map ALSA volume range

arcnmx:
Match [the approach that alsamixer uses](http://git.alsa-project.org/?p=alsa-utils.git;a=blob;f=alsamixer/volume_mapping.c;h=1c0d7c45e6686239464e1b0bbc8983ea57f3914f;hb=HEAD) to provide a more natural volume percentage:

> The mapping is designed so that the position in the interval is proportional to the volume as a human ear would perceive it (i.e., the position is the cubic root of the linear sample multiplication factor).  For controls with a small range (24 dB or less), the mapping is linear in the dB values so that each step has the same size visually.  Only for controls without dB information, a linear mapping of the hardware volume register values is used (this is the same algorithm as used in the old alsamixer).

### DIFF
--- a/src/print_volume.c
+++ b/src/print_volume.c
@@ -11,6 +11,7 @@
 #ifdef LINUX
 #include <alsa/asoundlib.h>
 #include <alloca.h>
+#include <math.h>
 #endif
 
 #if defined(__FreeBSD__) || defined(__DragonFly__)
@@ -111,11 +112,13 @@ void print_volume(yajl_gen json_gen, char *buffer, const char *fmt, const char *
 #endif
 
 #ifdef LINUX
+    const long MAX_LINEAR_DB_SCALE = 24;
     int err;
     snd_mixer_t *m;
     snd_mixer_selem_id_t *sid;
     snd_mixer_elem_t *elem;
     long min, max, val;
+    bool force_linear = false;
     int avg;
 
     if ((err = snd_mixer_open(&m, 0)) < 0) {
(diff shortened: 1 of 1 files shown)


stapelberg:
I’m not knowledgeable on this either :).

orestisf1993:
Potentially uneducated question: can we use GPL code with BSD-3?

Airblader:
You raise a good point. The answer to the best of my knowledge is no, we can't use it without having to relicense under the GPL.

Airblader:
In any case, unless alsa is dual licensed, I'd stick with the license of the project and not the header of the file.

I don't think "let me obfuscate the code" is a valid solution to working around the license issue. 

One option would be to contact the project and get clarification on the discrepancy in the header of the file. We certainly have to be careful here.

Airblader:
@stapelberg Should we close this due to the license concerns?

stapelberg:
What exactly is the licensing concern? While many source files in alsamixer carry a GPL-2+ header, the particular file volume_mapping.c carries an ISC header, which is compatible with our license.

Airblader:
Is it valid to pick the license of the particular file if it's not the license of the entire project? If so, then alright.

src/print_volume.c
```diff
@@ -159,16 +163,33 @@ void print_volume(yajl_gen json_gen, char *buffer, const char *fmt, const char *
...
+        snd_mixer_selem_get_playback_volume_range(elem, &min, &max);
+        snd_mixer_selem_get_playback_volume(elem, 0, &val);
+        force_linear = true;
+    } else {
+        snd_mixer_selem_get_playback_dB(elem, 0, &val);
+    }
+
+    /* Use linear mapping for raw register values or small ranges of 24 dB */
+    if (force_linear || max - min <= MAX_LINEAR_DB_SCALE * 100) {
+        if (max - min != 100) {
```
orestisf1993: what does this check do?

Airblader:
@arcnmx Could you address the comment @orestisf1993 posted? Then we could go ahead with this, I think.

src/print_volume.c
```diff
@@ -161,16 +165,37 @@ void print_volume(yajl_gen json_gen, char *buffer, const char *fmt, const char *
     }
 
     /* Get the volume range to convert the volume later */
-    snd_mixer_selem_get_playback_volume_range(elem, &min, &max);
-
     snd_mixer_handle_events(m);
-    snd_mixer_selem_get_playback_volume(elem, 0, &val);
-    if (max != 100) {
-        float avgf = ((float)val / max) * 100;
+    err = snd_mixer_selem_get_playback_dB_range(elem, &min, &max);
```
orestisf1993: I like
```c
/* Get the volume range to convert the volume later */
snd_mixer_handle_events(m);
err = snd_mixer_selem_get_playback_dB_range(elem, &min, &max) ||
      snd_mixer_selem_get_playback_dB(elem, 0, &val);
if (err < 0 || min >= max) {
    err = snd_mixer_selem_get_playback_volume_range(elem, &min, &max) ||
          snd_mixer_selem_get_playback_volume(elem, 0, &val);
    force_linear = true;
}
```
more because it is a bit more compact and will get the linear volume range if `snd_mixer_selem_get_playback_dB` fails but `snd_mixer_selem_get_playback_dB_range` not.

arcnmx: Tempted to use `?:` in that case to preserve the error code but that's a gnuism isn't it...

orestisf1993:
Thanks!

Issue #165, opened 2016-09-25
Monitor memory usage


lahwaacz:
The [story in the man page](https://github.com/i3/i3status/blob/8d2ef5f99b2ebc08a0a9d1f26492094692b1fc6b/man/i3status.man#L600-L611) is half user-specific and half false.

1. There are reasons why users might want to monitor memory usage. For example, I have 8 GB of memory and applications running most of the time in my session (e.g. browser) can easily take about 4 GB. Yet from time to time I need to compile a project which temporarily needs about 6 GB of memory. There is nothing easier than to look at the status bar to see if I need to close the browser or something.
2. The technical reasons regarding the difficulties of measuring memory usage don't apply anymore, at least on Linux, which does the hard work for us. i3status could just read the `MemTotal` and `MemAvailable` fields from `/proc/meminfo` and calculate a percentage. See [proc.txt](https://github.com/torvalds/linux/blob/1b5946a84d6eb096158e535bdb9bda06e7cdd941/Documentation/filesystems/proc.txt#L834-L961) in the kernel documentation.

stapelberg:
1. When I run work loads which exceed my available memory, the OOM killer kills processes (corresponding to tabs) as necessary. Is that not what you’re seeing on your machine? I’m using Chrome, but I’m assuming that Firefox with [e10s](https://wiki.mozilla.org/Electrolysis) enabled should behave the same way. From a philosophical standpoint, I still think that you shouldn’t need to monitor memory usage and just let your OOM killer do what you would do manually.

2. Interesting, thanks for pointing that out! https://blog.famzah.net/2014/09/24/memavailable-metric-for-linux-kernels-before-3-14-in-procmeminfo/ says MemAvailable was introduced in Linux 3.14.

Airblader:
For what it's worth, I support adding a memory output option. It's easily more relevant than displaying my ip address and I definitely think users may want to track it. Chrome may have an OOM but my private machine also has only 8 GB and running a VM and then some other things can cause this clash and often don't end nicely.

lahwaacz:
@stapelberg OK, two more points:

[...]

PR #270, opened 2018-03-20
initial support of memory-usage for linux

Stunkymonkey:
a lot of people are waiting for this feature.
so i implemented it for linux

### DIFF
--- a/i3status.c
+++ b/i3status.c
@@ -423,13 +423,10 @@ int main(int argc, char *argv[]) {
 
     cfg_opt_t memory_opts[] = {
         CFG_STR("format", "%used %free %available", CFGF_NONE),
-        CFG_STR("degraded_format_below_threshold", NULL, CFGF_NONE),
-        CFG_STR("degraded_threshold_type", "percentage_avail", CFGF_NONE),
-        CFG_FLOAT("degraded_low_threshold", 0, CFGF_NONE),
-        CFG_STR("critical_format_below_threshold", NULL, CFGF_NONE),
-        CFG_STR("critical_threshold_type", "percentage_avail", CFGF_NONE),
-        CFG_FLOAT("critical_low_threshold", 0, CFGF_NONE),
-        CFG_BOOL("use_available_memory", true, CFGF_NONE),
+        CFG_STR("format_degraded", NULL, CFGF_NONE),
+        CFG_STR("threshold_degraded", NULL, CFGF_NONE),
+        CFG_STR("threshold_critical", NULL, CFGF_NONE),
+        CFG_STR("memory_used_method", "classical", CFGF_NONE),
         CFG_CUSTOM_ALIGN_OPT,
         CFG_CUSTOM_COLOR_OPTS,
         CFG_CUSTOM_MIN_WIDTH_OPT,
@@ -743,7 +740,7 @@ int main(int argc, char *argv[]) {
 
             CASE_SEC("memory") {
                 SEC_OPEN_MAP("memory");
-                print_memory(json_gen, buffer, cfg_getstr(sec, "format"), cfg_getstr(sec, "degraded_format_below_threshold"), cfg_getstr(sec, "degraded_threshold_type"), cfg_getfloat(sec, "degraded_low_threshold"), cfg_getstr(sec, "critical_format_below_threshold"), cfg_getstr(sec, "critical_threshold_type"), cfg_getfloat(sec, "critical_low_threshold"), cfg_getbool(sec, "use_available_memory"));
+                print_memory(json_gen, buffer, cfg_getstr(sec, "format"), cfg_getstr(sec, "format_degraded"), cfg_getstr(sec, "threshold_degraded"), cfg_getstr(sec, "threshold_critical"), cfg_getstr(sec, "memory_used_method"));
                 SEC_CLOSE_MAP;
             }
 
--- a/i3status.c
+++ b/i3status.c
@@ -421,6 +421,22 @@ int main(int argc, char *argv[]) {
         CFG_CUSTOM_SEP_BLOCK_WIDTH_OPT,
         CFG_END()};
 
+    cfg_opt_t memory_opts[] = {
+        CFG_STR("format", "%used %free %available", CFGF_NONE),
+        CFG_STR("degraded_format_below_threshold", NULL, CFGF_NONE),
+        CFG_STR("degraded_threshold_type", "percentage_avail", CFGF_NONE),
+        CFG_FLOAT("degraded_low_threshold", 0, CFGF_NONE),
+        CFG_STR("critical_format_below_threshold", NULL, CFGF_NONE),
+        CFG_STR("critical_threshold_type", "percentage_avail", CFGF_NONE),
+        CFG_FLOAT("critical_low_threshold", 0, CFGF_NONE),
+        CFG_BOOL("use_available_memory", true, CFGF_NONE),
+        CFG_CUSTOM_ALIGN_OPT,
+        CFG_CUSTOM_COLOR_OPTS,
+        CFG_CUSTOM_MIN_WIDTH_OPT,
+        CFG_CUSTOM_SEPARATOR_OPT,
+        CFG_CUSTOM_SEP_BLOCK_WIDTH_OPT,
+        CFG_END()};
+
     cfg_opt_t usage_opts[] = {
         CFG_STR("format", "%usage", CFGF_NONE),
         CFG_STR("format_above_threshold", NULL, CFGF_NONE),
@@ -490,6 +506,7 @@ int main(int argc, char *argv[]) {
         CFG_SEC("tztime", tztime_opts, CFGF_TITLE | CFGF_MULTI),
         CFG_SEC("ddate", ddate_opts, CFGF_NONE),
         CFG_SEC("load", load_opts, CFGF_NONE),
+        CFG_SEC("memory", memory_opts, CFGF_NONE),
         CFG_SEC("cpu_usage", usage_opts, CFGF_NONE),
         CFG_END()};
 
@@ -721,6 +738,12 @@ int main(int argc, char *argv[]) {
             CASE_SEC("load") {
                 SEC_OPEN_MAP("load");
                 print_load(json_gen, buffer, cfg_getstr(sec, "format"), cfg_getstr(sec, "format_above_threshold"), cfg_getfloat(sec, "max_threshold"));
+                SEC_CLOSE_MAP;
+            }
+
+            CASE_SEC("memory") {
+                SEC_OPEN_MAP("memory");
+                print_memory(json_gen, buffer, cfg_getstr(sec, "format"), cfg_getstr(sec, "degraded_format_below_threshold"), cfg_getstr(sec, "degraded_threshold_type"), cfg_getfloat(sec, "degraded_low_threshold"), cfg_getstr(sec, "critical_format_below_threshold"), cfg_getstr(sec, "critical_threshold_type"), cfg_getfloat(sec, "critical_low_threshold"), cfg_getbool(sec, "use_available_memory"));
                 SEC_CLOSE_MAP;
             }
 
--- a/man/i3status.man
+++ b/man/i3status.man
@@ -55,6 +55,7 @@ order += "wireless wlan0"
 order += "ethernet eth0"
 order += "battery 0"
 order += "cpu_temperature 0"
+order += "memory"
 order += "load"
 order += "tztime local"
 order += "tztime berlin"
@@ -111,6 +112,10 @@ load {
 cpu_temperature 0 {
 	format = "T: %degrees °C"
 	path = "/sys/devices/platform/coretemp.0/temp1_input"
+}
+
+memory {
+        format = "%used"
 }
 
 disk "/" {
@@ -423,6 +428,49 @@ starting from %cpu0. This feature is currently not supported in FreeBSD.
 *Example degraded_threshold*: +25+
 
 *Example format_above_degraded_threshold*: +Warning above degraded threshold: %usage+
+
+=== Memory
+
+Gets the memory usage from system.
+On Linux, the information is taken from +/proc/meminfo+.
+
+These values can also be expressed in percentages with the +percentage_used+,
++percentage_free+ and +percentage_shared+ formats.
+
+If a +critical_low_threshold+ is defined, the output will be colored using
++color_bad+. The unit of this threshold is defined by the unit given in
++critical_threshold_type+, which can be one of "bytes_free", "bytes_avail",
+"percentage_free" and "percentage_avail". Additionally, the former two can be
+prefixed with one of "k", "m", "g" or "t" to change the exact unit.
+For example, setting critical_low_threshold to 2 and threshold_type to "gbytes_avail"
+causes available memory below 2 GiB to be colored with +color_bad+. The
+defaults are "percentage_avail" with a threshold of 0.
+Furthermore, the format used when the threshold is reached can be customized
+using the option +critical_format_below_threshold+.
+Same applies to +degraded_low_threshold+, +degraded_threshold_type+ and
++degraded_format_below_threshold+ using +color_degraded+.
+
+The most common one is:
+"used memory" = "total memory" - "free memory" - "buffers" - "cache"
+This is the default in i3status. Some other programs use
+"used memory" = "total memory" - "available memory"
+You can disable this behavior using +use_available_memory+.
+
+*Example order*: +memory+
+
+*Example format*: +%free %available (%used)/ %total+
+
+*Example format*: +used %percentage_used , free %percentage_free, shared %percentage_shared+
+
+*Example degraded_low_threshold*: +10+
+
+*Example degraded_threshold_type*: +percentage_free+
+
+*Example critical_low_threshold*: +5+
+
+*Example critical_format_below_threshold*: +Warning: %percentage_free+
+
+*Example use_available_memory: +false+
 
 === Load
 
--- a/man/i3status.man
+++ b/man/i3status.man
@@ -116,6 +116,8 @@ cpu_temperature 0 {
 
 memory {
         format = "%used"
+        threshold_degraded = "10%"
+        format_degraded = "MEMORY: %free"
 }
 
 disk "/" {
@@ -197,6 +199,7 @@ double-quote (""") characters need to be replaced with "`&amp;`", "`&lt;`",
 for generated content (e.g. wireless ESSID, time).
 
 *Example configuration*:
+
 -------------------------------------------------------------
 general {
     output_format = "xmobar"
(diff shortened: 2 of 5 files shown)


man/i3status.man
```diff
@@ -55,6 +55,7 @@ order += "wireless wlan0"
 order += "ethernet eth0"
 order += "battery 0"
 order += "cpu_temperature 0"
+order += "cpu_usage"
```
bebehei: the memory block is missing in here!?

src/print_mem.c
```diff
@@ -0,0 +1,181 @@
...
+    const char *selected_format = format;
+    const char *walk;
+    bool colorful_output = false;
+
+    double totalRam = 0.0;
+    double freeRam = 0.0;
+    double usedRam = 0.0;
+
+    FILE *file = fopen(filename, "r");
+    if (file != NULL) {
```
bebehei: `if (file)`

src/print_mem.c
```diff
@@ -0,0 +1,181 @@
...
+    char *outwalk = buffer;
+/* Get memory */
+
+#if defined(linux)
+    const char *selected_format = format;
+    const char *walk;
+    bool colorful_output = false;
+
+    double totalRam = 0.0;
+    double freeRam = 0.0;
```
bebehei: Please use integers and initialize them to `-1`. This gives you the possibility to compare the integer to `-1` and be sure, that it's still the default value.

src/print_mem.c
```diff
@@ -0,0 +1,181 @@
...
+
+    if (colorful_output)
+        END_COLOR;
+
+    *outwalk = '\0';
+    OUTPUT_FULL_TEXT(buffer);
+
+    return;
+error:
+#endif
```
bebehei: As there is no support for systems other than Linux, a more explanatory message encapsuled within `#else` might be a good choice.

src/print_mem.c
```diff
@@ -0,0 +1,181 @@
...
+static const char *const si_symbols[MAX_EXPONENT + 1] = {"", "k", "M", "G", "T"};
+static const char *const custom_symbols[MAX_EXPONENT + 1] = {"", "K", "M", "G", "T"};
+
+double get_double(const char *str) {
+    /* First skip non-digit characters */
+    /* Special case to handle negative numbers and the `+` sign */
+    while (*str && !(isdigit(*str) || ((*str == '-' || *str == '+') && isdigit(*(str + 1)))))
+        str++;
+
+    /* The parse to a double */
```
bebehei: CS 1. sem

src/print_mem.c
```diff
@@ -0,0 +1,181 @@
...
+        END_COLOR;
+
+    *outwalk = '\0';
+    OUTPUT_FULL_TEXT(buffer);
+
+    return;
+error:
+#endif
+    OUTPUT_FULL_TEXT("can't read memory");
+    (void)fputs("i3status: Cannot read system memory using /proc/meminfo\n", stderr);
```
bebehei: I guess, `(void)` isn't necessary here.

Airblader:
Hi, thanks for the PR. Unfortunately, I'm going to have to close this because this is not a feature we are looking to add in i3status. You can find discussions on this in several closed issues by searching for keywords like "memory".

For future reference, we would encourage you to open an issue first to discuss nee features as to avoid potentially unnecessary work. 

That said, you and anyone else can of course apply this patch locally, so thanks for sharing.

Airblader:
Reopening for now as discussed in #165

src/print_mem.c
Stunkymonkey: i used doubles, because outputting numbers with commas was easier this way
when i switch it, should i use `size_t` instead?

man/i3status.man
```diff
@@ -424,6 +433,52 @@ starting from %cpu0. This feature is currently not supported in FreeBSD.
...
+whose type can be specified via the "prefix_type" option. Three sets of
+prefixes are available:
+
+binary::
+IEC prefixes (Ki, Mi, Gi, Ti) represent multiples of powers of 1024.
+This is the default.
+decimal::
+SI prefixes (k, M, G, T) represent multiples of powers of 1000.
+custom::
+The custom prefixes (K, M, G, T) represent multiples of powers of 1024.
```
lahwaacz: Typo: except for k/K, this is the same as the SI prefixes two lines above.

What is the difference between "binary" and "custom" since they appear to use the same base?

src/print_mem.c
lahwaacz: You can't initialize a `size_t` variable to `-1`. You can still initialize the `double` variables to a negative value with the same effect as @bebehei had in mind.

src/print_mem.c
```diff
@@ -0,0 +1,181 @@
...
+
+    FILE *file = fopen(filename, "r");
+    if (file != NULL) {
+        char line[128];
+        while (fgets(line, sizeof line, file) != NULL) {
+            if (strncmp("MemTotal:", line, strlen("MemTotal:")) == 0) {
+                totalRam = get_double(line + strlen("MemTotal:"));
+            }
+            if (strncmp("MemAvailable:", line, strlen("MemAvailable:")) == 0) {
+                freeRam = get_double(line + strlen("MemAvailable:"));
```
lahwaacz: So you're just reading `MemTotal` and `MemAvailable` and using it for all calculations. Note that the `free` utility shows different info - from `free(1)`:
```
       total  Total installed memory (MemTotal and SwapTotal in /proc/meminfo)

       used   Used memory (calculated as total - free - buffers - cache)

       free   Unused memory (MemFree and SwapFree in /proc/meminfo)

       shared Memory used (mostly) by tmpfs (Shmem in /proc/meminfo)

       buffers
              Memory used by kernel buffers (Buffers in /proc/meminfo)

       cache  Memory  used  by  the  page  cache  and slabs (Cached and SReclaimable in
              /proc/meminfo)

       buff/cache
              Sum of buffers and cache

       available
              Estimation of how much memory is available for starting new applications,
              without  swapping.  Unlike the data provided by the cache or free fields,
              this  field  takes  into  account  page  cache  and  also  that  not  all
              reclaimable  memory  slabs  will  be  reclaimed due to items being in use
              (MemAvailable in /proc/meminfo, available on kernels  3.14,  emulated  on
              kernels 2.6.27+, otherwise the same as free)
```
Note that `total - free - buffers - cache` and `total - available` are typically very different, they are off by about 400 MiB on my system right now.

I think it's worth showing (or rather, making it possible to let the user show) the same values as `free`. All the needed info is in `/proc/meminfo` and adding more fields to the config is easy.

Stunkymonkey: should i implement all the variables you named?
so the user could choose which one he wants?

lahwaacz: Maybe `buffers` and `cache` would not have to be visible directly, but they should be used to calculate the `used` memory like free does it.

Stunkymonkey:
sorry had an exam today, that's why i was not responding.

src/print_mem.c
Stunkymonkey: what about `shared`?

man/i3status.man
Stunkymonkey: maybe some people want to use the `binary_base`, but prefer to have `GB` instead of `GiB`

src/print_mem.c
Stunkymonkey: i would not like to use `integers`, because linux itself uses `long` [source](https://github.com/torvalds/linux/blob/master/fs/proc/meminfo.c)

bebehei: Well, long is just a "longer" integer :joy:  It's not a floating point type. But anyways, as @lahwaacz noted, setting the `double` initially to a negative value, might be the best.

man/i3status.man
lahwaacz: Well, that's one completely unintuitive meaning of "custom". But since that's the same as for the "disk" i3status module, this can be dismissed.

i3status.c
```diff
@@ -421,6 +421,20 @@ int main(int argc, char *argv[]) {
         CFG_CUSTOM_SEP_BLOCK_WIDTH_OPT,
         CFG_END()};
 
+    cfg_opt_t memory_opts[] = {
+        CFG_STR("format", "%used %free %available", CFGF_NONE),
+        CFG_STR("format_above_threshold", NULL, CFGF_NONE),
+        CFG_STR("prefix_type", "binary", CFGF_NONE),
+        CFG_STR("threshold_type", "percentage_avail", CFGF_NONE),
+        CFG_BOOL("calc_used_with_available", false, CFGF_NONE),
```
Stunkymonkey: we should discuss what is the default for the `%used` variable

Stunkymonkey:
the Linux part is done and ready for reviewing

eplanet:
First, thanks for yours efforts. One little remark, I don't see any unit test, it would probably be easy to introduce one if the path to `/proc/meminfo` is parameterized. That would be great if you could add that. Thanks!

man/i3status.man
Airblader: Let's please remove this `custom` stuff. We want to display the correct unit name. If a user doesn't like that, they'll just need to use a different solution. Right now this option is confusing at best.

man/i3status.man
```diff
@@ -424,6 +429,58 @@ starting from %cpu0. This feature is currently not supported in FreeBSD.
...
+"bytes_free", "bytes_avail", "percentage_free", or "percentage_avail", where
+the former two can be prepended by a generic prefix (k, m, g, t) having
+prefix_type. So, if you configure low_threshold to 2, threshold_type to
+"gbytes_avail", and prefix_type to "binary", and the remaining available memory
+space is below 2 GiB, it will be colored bad. If not specified, threshold_type
+is assumed to be "percentage_avail" and low_threshold to be set to 0, which
+implies no coloring at all. You can customize the output format when below
+low_threshold with format_below_threshold.
+
+In Unix calculation the used memory is done in various way. The most used one is:
```
Airblader: This sentence doesn't sound right.

man/i3status.man
```diff
@@ -424,6 +429,58 @@ starting from %cpu0. This feature is currently not supported in FreeBSD.
...
+
+*Example format*: +%free %available (%used)/ %total+
+
+*Example format*: +%percentage_used used, %percentage_free free+
+
+*Example prefix_type*: +custom+
+
+*Example low_threshold*: +5+
+
+*Example format_below_threshold*: +Warning: %percentage_free+
```
Airblader: According to the code above, there's a `format_above_threshold`, not a `format_below_threshold`.

src/print_mem.c
```diff
@@ -0,0 +1,221 @@
...
+/*
+ * Prints the given amount of bytes in a human readable manner.
+ *
+ */
+static int print_bytes_human(char *outwalk, uint64_t bytes, const char *prefix_type) {
+    if (strcasecmp(prefix_type, "decimal") == 0) {
+        return format_bytes(outwalk, bytes, DECIMAL_BASE, si_symbols);
+    } else if (strcasecmp(prefix_type, "custom") == 0) {
+        return format_bytes(outwalk, bytes, BINARY_BASE, custom_symbols);
+    } else {
```
Airblader: Please use specific cases only and error out on invalid configurations.

src/print_mem.c
```diff
@@ -0,0 +1,221 @@
...
+            if (strncmp("Buffers:", line, strlen("Buffers:")) == 0) {
+                buffers = get_long(line + strlen("Buffers:"));
+            }
+            if (strncmp("Cached:", line, strlen("Cached:")) == 0) {
+                cached = get_long(line + strlen("Cached:"));
+            }
+            if (strncmp("Shmem:", line, strlen("Shmem:")) == 0) {
+                sharedRam = get_long(line + strlen("Shmem:"));
+            }
+            if (totalRam > 0 && freeRam > 0 && availableRam > 0 && buffers > 0 && cached > 0 && sharedRam > 0) {
```
Airblader: Shouldn't these be `>=`?

src/print_mem.c
```diff
@@ -0,0 +1,221 @@
...
+    long totalRam = -1;
+    long freeRam = -1;
+    long availableRam = -1;
+    long usedRam = -1;
+    long sharedRam = -1;
+    long cached = -1;
+    long buffers = -1;
+
+    FILE *file = fopen(filename, "r");
+    if (file) {
```
Airblader: If we invert this to

```
if (!file) {
    goto error;
}
```

we can avoid the indentation.

src/print_mem.c
```diff
@@ -0,0 +1,221 @@
...
+    if (calc_used_with_available) {
+        usedRam = totalRam - availableRam;
+    } else {
+        usedRam = totalRam - freeRam - buffers - cached;
+    }
+
+    if (low_threshold > 0 && below_threshold(totalRam, usedRam, prefix_type, threshold_type, low_threshold)) {
+        START_COLOR("color_bad");
+        colorful_output = true;
+        if (format_above_threshold != NULL)
```
Airblader: If we're below the threshold, we use an above_threshold format? That sounds wrong.

src/print_mem.c
```diff
@@ -0,0 +1,221 @@
...
+                return false;
+        }
+        return free < low_threshold * factor;
+    }
+    return false;
+}
+
+void print_memory(yajl_gen json_gen, char *buffer, const char *format, const char *format_above_threshold, const char *prefix_type, const char *threshold_type, const float low_threshold, const bool calc_used_with_available) {
+    char *outwalk = buffer;
+/* Get memory */
```
bebehei: Albeit you explained me, that clang-format requires this comment to be unindented, I still think it's looking very ill. As this comment doesn't add any value in this position, I'd think removing it would be best.

src/print_mem.c
```diff
@@ -0,0 +1,221 @@
...
+static const char filename[] = "/proc/meminfo";
+
+static const char *const iec_symbols[MAX_EXPONENT + 1] = {"", "Ki", "Mi", "Gi", "Ti"};
+static const char *const si_symbols[MAX_EXPONENT + 1] = {"", "k", "M", "G", "T"};
+static const char *const custom_symbols[MAX_EXPONENT + 1] = {"", "K", "M", "G", "T"};
+
+long get_long(const char *str) {
+    /* First skip non-digit characters */
+    /* Special case to handle negative numbers and the `+` sign */
+    while (*str && !(isdigit(*str) || ((*str == '-' || *str == '+') && isdigit(*(str + 1)))))
```
Airblader: Why exactly do we need this special handling? `man strtol` says

>       The  string  may  begin  with  an arbitrary amount of white space (as determined by iss‐
>       pace(3)) followed by a single optional '+' or '-' sign.

Stunkymonkey: Oh I didn't know that.

Stunkymonkey:
@Airblader & @bebehei i fixed everything according to your comments

i3status.c
```diff
@@ -421,6 +421,20 @@ int main(int argc, char *argv[]) {
         CFG_CUSTOM_SEP_BLOCK_WIDTH_OPT,
         CFG_END()};
 
+    cfg_opt_t memory_opts[] = {
+        CFG_STR("format", "%used %free %available", CFGF_NONE),
+        CFG_STR("format_above_threshold", NULL, CFGF_NONE),
```
Airblader: This still says above.

man/i3status.man
```diff
@@ -424,6 +429,58 @@ starting from %cpu0. This feature is currently not supported in FreeBSD.
 
 *Example format_above_degraded_threshold*: +Warning above degraded threshold: %usage+
 
+=== Memory
+
+Gets used, free and total amount of memory on the current system. Gets the Memory usage
```
Airblader: Gets the Memory usage from […] (Linux)

→

On Linux, the information is taken from +/proc/meminfo+.

man/i3status.man
```diff
@@ -424,6 +429,58 @@ starting from %cpu0. This feature is currently not supported in FreeBSD.
...
+"bytes_free", "bytes_avail", "percentage_free", or "percentage_avail", where
+the former two can be prepended by a generic prefix (k, m, g, t) having
+prefix_type. So, if you configure low_threshold to 2, threshold_type to
+"gbytes_avail", and prefix_type to "binary", and the remaining available memory
+space is below 2 GiB, it will be colored bad. If not specified, threshold_type
+is assumed to be "percentage_avail" and low_threshold to be set to 0, which
+implies no coloring at all. You can customize the output format when below
+low_threshold with format_below_threshold.
+
+In Unix calculating the used memory is done in different ways. The most commonly one is:
```
Airblader: s/commonly/common/

man/i3status.man
```diff
@@ -424,6 +429,58 @@ starting from %cpu0. This feature is currently not supported in FreeBSD.
...
+"used memory" = "total memory" - "free memory" - "buffers" - "cache" (which is
+default in i3status) some programms use: "used memory" = "total memory" -
```
Airblader: This is a new sentence, so it should be capitalized and started after a period:

> The most common one is:
>   "used memory" = "total memory" - "free memory" - "buffers" - "cache"
> This is the default in i3status. Some other programs use
>   "used memory" = "total memory" - "available memory"
> You can enable this behavior using +calc_used_with_available+.

man/i3status.man
```diff
@@ -424,6 +429,58 @@ starting from %cpu0. This feature is currently not supported in FreeBSD.
...
+"available memory". You can turn this on with the "calc_used_with_available".
```
Airblader: In fact, let's please rename this option to `use_available_memory`

Airblader:
I just realized that the "custom" was probably added because the disk info also has this option. @stapelberg Do you ("still") think this is a good idea? I'd actually prefer leaving such an option out and always use the correct unit. In particular "custom" for "use the wrong name for this base" is a really odd naming.

stapelberg:
I think it’s reasonable to support 1000 and 1024 based units for disks, where manufacturers frequently use the 1000 based ones.

For RAM, I think 1024 is prevalent enough that we should only display those.

man/i3status.man
lahwaacz: Additionally, there should be full stops after each sentence.

Stunkymonkey:
should i remove all prefix types and use binary as default?

Airblader:
Yes, please. Thanks!

man/i3status.man
```diff
@@ -424,6 +429,59 @@ starting from %cpu0. This feature is currently not supported in FreeBSD.
...
 
+=== Memory
+
+Gets the Memory usage from system.
+On Linux, the information is taken from +/proc/meminfo+.
+
+These values can also be expressed in percentages with the +percentage_used+,
++percentage_free+ and +percentage_shared+ formats.
+
+Byte sizes are presented in a human readable format using a set of prefixes
```
Airblader: All of this documentation on the prefixes needs to be removed as well.

Stunkymonkey: updated

Stunkymonkey:
updated again

man/i3status.man
```diff
@@ -424,6 +429,45 @@ starting from %cpu0. This feature is currently not supported in FreeBSD.
 
 *Example format_above_degraded_threshold*: +Warning above degraded threshold: %usage+
 
+=== Memory
+
+Gets the Memory usage from system.
```
Airblader: s/Memory/memory

man/i3status.man
```diff
@@ -424,6 +429,45 @@ starting from %cpu0. This feature is currently not supported in FreeBSD.
...
++percentage_free+ and +percentage_shared+ formats.
+
+It is possible to define a low_threshold that causes the memory text to be
+displayed using color_bad. The low_threshold type can be of threshold_type
+"bytes_free", "bytes_avail", "percentage_free", or "percentage_avail", where
+the former two can be prepended by a generic prefix (k, m, g, t) having
+prefix_type. So, if you configure low_threshold to 2, threshold_type to
+"gbytes_avail", and prefix_type to "binary", and the remaining available memory
+space is below 2 GiB, it will be colored bad. If not specified, threshold_type
+is assumed to be "percentage_avail" and low_threshold to be set to 0, which
```
Airblader: The documented default doesn't seem to agree with the actual default. There's also a couple of IMHO hard to read things in here as well as still leftover references to the removed options. Can we rephrase the entire paragraph like this?

> If a +low_threshold+ is defined, the output will be colored using +color_bad+.
> The unit of this threshold is defined by the unit given in +threshold_type+,
> which can be one of "bytes_free", "bytes_avail", "percentage_free" and 
> "percentage_avail". Additionally, the former two can be prefixed with one of
> "k", "m", "g" or "t" to change the exact unit.
> For example, setting low_threshold to 2 and threshold_type to "gbytes_avail"
> causes available memory below 2 GiB to be colored with +color_bad+. The 
> defaults are "percentage_avail" with a threshold of 0.
> Furthermore, the format used when the threshold is reached can be customized
> using the option +format_below_threshold+.

i3status.c
```diff
@@ -421,6 +421,19 @@ int main(int argc, char *argv[]) {
         CFG_CUSTOM_SEP_BLOCK_WIDTH_OPT,
         CFG_END()};
 
+    cfg_opt_t memory_opts[] = {
+        CFG_STR("format", "%used %free %available", CFGF_NONE),
+        CFG_STR("format_below_threshold", NULL, CFGF_NONE),
+        CFG_STR("threshold_type", "percentage_avail", CFGF_NONE),
+        CFG_BOOL("use_available_memory", true, CFGF_NONE),
+        CFG_FLOAT("low_threshold", 5, CFGF_NONE),
```
Airblader: Should be 0, right?

Stunkymonkey:
all things done

i3status.c
```diff
@@ -724,6 +738,12 @@ int main(int argc, char *argv[]) {
                 SEC_CLOSE_MAP;
             }
 
+            CASE_SEC("memory") {
+                SEC_OPEN_MAP("memory");
+                print_memory(json_gen, buffer, cfg_getstr(sec, "format"), cfg_getstr(sec, "format_above_threshold"), cfg_getstr(sec, "threshold_type"), cfg_getfloat(sec, "low_threshold"), cfg_getbool(sec, "use_available_memory"));
```
ixjlyons: This should be `format_below_threshold`

Stunkymonkey: forgot this one
is fixed

bebehei:
Does anyone have an objection against also having a `color_degraded` setting in the memory? I'd really appreciate this.

Airblader:
I don't mind adding that.

Stunkymonkey:
ok i will add it :+1:

Stunkymonkey:
have a look at it :smile: 
specially on the man page and the variable-naming

Stunkymonkey:
if someone wants to start on a netbsd solution [here](https://github.com/brndnmtthws/conky/blob/6c4c28672f3a09a412941f3d58e40f6391389bd6/src/netbsd.cc#L118-L146)

bebehei:
@Airblader are you at the GPN? If so, let's meet, please! @Stunkymonkey and myself are there.

Airblader:
Unfortunately I am not, sorry :-)

Stunkymonkey:
> I just want to double-check with everyone that it's a go from your perspective?

Yes go for it

Stunkymonkey:
ping

Airblader:
I haven't forgotten, don't worry. :-) Thanks everyone, for contributing and also for the patience!

PR #273, opened 2018-03-28
Fix NULL value on OpenBSD when there is no acpitz0

winks:
Using %degrees on OpenBSD resulted in garbage values if sysctl doesn't
know about any cpu temperature sensors.

### DIFF
--- a/src/print_cpu_temperature.c
+++ b/src/print_cpu_temperature.c
@@ -96,6 +96,7 @@ static int read_temperature(char *thermal_zone, temperature_t *temperature) {
     struct sensor sensor;
     size_t sdlen, slen;
     int dev, numt, mib[5] = {CTL_HW, HW_SENSORS, 0, 0, 0};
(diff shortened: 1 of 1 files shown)


Airblader:
I think it'd be better to just first initialize the structure to some empty value and then have the logic run.

su8:
```c
memset(&sensordev, 0, sizeof(struct sensordev));
memset(&sensor, 0, sizeof(struct sensor));
```

su8:
You can also add

On line [124](https://github.com/winks/i3status/blob/93c82ffc25302fea1f12cfda5ac3b6bcdc2f189b/src/print_cpu_temperature.c#L124) you can add simple switch

```c
case SENSOR_TEMP:
  // do stuff
```

Airblader:
I take that as you not pursuing to make changes to the PR, correct? Then I'll close it. Maybe someone else will want to pick it up. :-)

PR #275, opened 2018-04-06
Fix NULL value on OpenBSD when there is no acpitz0

winks:
Using %degrees on OpenBSD resulted in garbage values if sysctl doesn't
know about any cpu temperature sensors.


winks:
Apologies, I misread something in #273 - and thought I'd had to change a lot and only got to look at it in depth just now

what @su8 suggested didn't work (inside the OpenBSD ifdef) but I think this could benefit all codepaths to have a defined 0 value.

Tested compile and run on OpenBSD 6.2 and Ubuntu 16.02.

Airblader:
Thanks!

Issue #276, opened 2018-04-07
Why using cat to read the file content ?


su8:
You could use slurp mode instead `cat` 

https://github.com/i3/i3status/blob/199a6a27136a5ddb3d813185f0f8cf9185072f08/travis/run-tests.pl#L22

Can be replaced with:

```perl
        open($fh, '<:encoding(iso-8859-1)', "@_/expected_output.txt")
          or die "Could not open file '@_/expected_output.txt' $!";
        local $/ = undef; # <--- slurp mode
        $refres = <$fh>;
        close($fh);
```

---

Updated the encoding as it failed with UTF-8, it took me awhile to find out which encoding contains the degree symbol.

su8:
You could drop the encoding as it encodes the ansi escapes the temperature symbols and returs different string that the wanted one.

stapelberg:
Could you send a pull request for this please? Also, we shouldn’t need to use iso-8859-1 anywhere. If our tests don’t use UTF-8, that’s another bug to fix right there :)

PR #277, opened 2018-04-07
Fixes https://github.com/i3/i3status/issues/276

su8:


Airblader:
Why is this an improvement? It requires more code and more explicit handling of the encoding. What's wrong with cat?

su8:
`cat foo`, there is no check whether this command succeeded or not.

The improvement is that we use file handles instead spawning a new process and check whether the file exists, if not simply **die** with pre-defined message.

su8:
If shorter version is required, then check the cat return value 

```perl

die "There is no @_/expected_output.txt" unless $? == cats_return_value_equals_ok

```

Airblader:
I appreciate you filing the PR, but I think I'll close it for now because I don't think the tradeoff is very good here. Thanks for understanding!

PR #174, opened 2016-11-02
Add option to display keyboard layout

dopefishh:
This PR adds functionality to show current active keyboard layout in the status line. This adds the dependency of libX11.

I've ran `clang-format` on all my new code but not on the old existing code.

PR #279, opened 2018-04-13
Add xkb layout option

baruch:
This is a preliminary work I did for myself, it is still incomplete with all the features I want but the pull request is intended to gauge interest in this work.

I still want to add a thread to get events on layout change so the display  update will be immediate upon layout change, and also probably add some options to display the best name or language.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ addons:
     - libasound2-dev
     - libpulse-dev
     - libnl-genl-3-dev
+    - xserver-xorg-dev
     - asciidoc
     # xmlto pulls in a2x which we need for generating manpages from asciidoc.
     - xmlto
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,7 @@ LIBS+=-lyajl
 LIBS+=-lpulse
 LIBS+=-lm
 LIBS+=-lpthread
+LIBS+=-lX11 -lxkbfile
 
 ifeq ($(wildcard .git),)
   # not in git repository
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ i3status has the following dependencies:
   * libcap2-bin (for getting network status without root permissions)
   * asciidoc (only for the documentation)
   * libpulse-dev (for getting the current volume using PulseAudio)
+  * xserver-xorg-dev (for xkb layout)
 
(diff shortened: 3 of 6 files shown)


baruch:
I've pushed an update to add the new dependency to travis for it to build, hopefully github/travis update the branch and retest soon.

Airblader:
#174

clarcharr:
@Airblader If you honestly think that such an option is not useful because you can just "type and know what keyboard layout you're using," I honestly don't think you've ever used more than two keyboard layouts in rotation. It's a pain in the ass and I know people who have 3-4 keyboard layouts and it may be confusing to figure out which you're on, because a lot of layouts share keys. You basically have to systematically go through every layout and type a known phrase to see if it works, and that's something that can easily be solved by listing the current layout in the status bar.

Why are you unwilling to accept this as a feature when so many people are willing to implement it for you?

baruch:
@clarcharr There is actually an easy way to solve this that doesnt touch the status bar, I've resorted to using gxkb which can also show the flag so is easier to visually find it. It does only show up on one status bar rather than both (I have dual-screen) but it works.

@Airblader  To avoid the repeated work to implement this feature that is obviously needed it might be a good idea to add the solution to the documentation.

Airblader:
> Why are you unwilling to accept this as a feature when so many people are willing to implement it for you?

While I can understand the frustration of implementing something which ends up being rejected, we don't (and don't want to) base feature decisions on whether there's code for it already. Such decisions are more complicated as a feature also comes with forever-lasting maintenance effort. i3 as a whole, and specifically i3status, is rather conservative in the feature selection also for reasons such as keeping configuration overhead and mental complexity to a minimum.

We generally recommend discussing a feature before writing the code exactly to prevent this kind of issue.

> To avoid the repeated work to implement this feature that is obviously needed it might be a good idea to add the solution to the documentation.

We already have documentation for how to replace i3status with another statusline generator or how to enrich i3status' output with custom information. Do you have a specific idea of where and how to document this? I'm not much of a fan of documenting every feature i3status *doesn't* have, to be honest… :-)

On a side note: if someone comes up with a PR without checking and finding one of the multiple issues and PRs about this, I think chances are they won't see that documentation either. So there's a bit of a catch 22 there.

PR #284, opened 2018-05-29
Add a general and per-module default color configuration

Gravemind:
Add a new `color_default` configuration to emit a per-module default color.

If left unspecified, you get the old behavior: the non-good/degraded/bad state does not emit any color, and so you get your *bar program*'s default color.

If `color_default` is specified, in `general` for all modules, or per-module in their respective sections, then the module will emit this color when in non-good/degraded/bad state.

Any color (`color_default`, `color_good` etc.) can be set to an empty string to disable the color, reverting to the *bar program* default color.

Notes on the source code changes:

- Replaced `START_COLOR`/`END_COLOR` with a variable `outcolor` (similar to `outwalk`), typed with a enum `output_color_t`. I think it's easier to play with the color now, and it makes it clearer it's for the entire module (as opposed to starting a color mid-sentence).
- This effectively adds a 4th state to good/bad/degraded. I thought about merging default and good state (e.g. good state by default), but some modules seem to differentiate between the default and the good state.

### DIFF
--- a/i3status.c
+++ b/i3status.c
@@ -37,9 +37,12 @@
 #define CFG_CUSTOM_ALIGN_OPT \
     CFG_STR_CB("align", NULL, CFGF_NONE, parse_align)
 
+#define CFG_CUSTOM_DEFAULT_COLOR_OPT \
+    CFG_STR("color_default", NULL, CFGF_NONE)
+
 #define CFG_COLOR_OPTS(good, degraded, bad)             \
-    CFG_STR("color_good", good, CFGF_NONE)              \
-    ,                                                   \
+    CFG_CUSTOM_DEFAULT_COLOR_OPT,                       \
+        CFG_STR("color_good", good, CFGF_NONE),         \
         CFG_STR("color_degraded", degraded, CFGF_NONE), \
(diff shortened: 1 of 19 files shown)


Gravemind:
For what it's worth, I still use it, and, anyway, I don't mind maintaining it just for me...

I thought my last description/commit message was not great, so I reworded it to something better (I hope):

---

The significant code change was to replace the `START_COLOR` and `END_COLOR` macros by an enum and a variable `outcolor` used similarly to `outwalk` (see `OUTPUT_FULL_TEXT`).

A few improvements have been made from there:

1. allow any `color_*` config to be set to an empty string to disable it (i.e. force remove the color from the output) (also useful for 2.).

2. add a `color_default` config to general and to all modules/sections, so the default/normal color can be customized everywhere (color of `tztime`, "normal" color of `memory`, etc...). Unspecified by default, so same as old behavior (no color output).

3. fix per-section color emission for `output_format` other than `i3bar`.

Gravemind:
I changed the title and description again.

FWIW I still thinks this is a good change, for users, and for the code.

I am open to suggestions. If you think this is outside the scope of i3status, I'm ok with closing this PR.

PR #285, opened 2018-06-02
use format_placeholder from i3

Stunkymonkey:
its now possible to have percentage before and after a variable. except
for the date. But percentage with dates does not make much sense to me, so
i skipped it.


Airblader:
Thanks for the PR. Did you miss my comment on the issue? We have a function in libi3 that handles replacements. I think we should add this function to i3status as well and use it in all places here instead.

> Note that in libi3 we have a format_placeholders. It might make sense to think about reusing this.

Stunkymonkey:
oh. i missed your comment...

Stunkymonkey:
so should i continue to implement the `format_placeholders` ?

Airblader:
Yeah, I think it'd be great if you could do that. I don't think there's a way to avoid duplicating the function from libi3 to i3status, but that's OK, we can just copy-paste it. It's better to have it duplicated once across projects than duplicating the logic multiple times within i3status.

Stunkymonkey:
No need to review this commit. Still needs lot of work.
But it works for `print-mem.c`.

Stunkymonkey:
the `START_COLOR` and `END_COLOR` are not yet changed.
but the rest of `print-memory.c` is done.

@Airblader is it implemented how you imagined it?

should i continue to start working on all other parts?

Airblader:
Yes, this looks much better I think. Please go ahead :-)

Stunkymonkey:
the `src/print_cpu_usage.c` is still missing and a bit tricky because of the `%cpu<number>`
rest is done

eplanet:
That looks like a very complicated change to in the end just being able to print a percentage sign. I'm afraid we're going to introduce bugs for a feature that IMHO adds very limited value to `i3status`.

Stunkymonkey:
for the percentage sign i suggested only the first commit.

Airblader:
I agree that for the sole purpose of fixing the bug it's a big change, but the point here is that we had this functionality of formatting the output duplicated all over the place. If you have to fix one bug in more than one place, it's a good sign that it's time for some refactoring.

Some of the main complexity added here is the formatting function itself, which is a copy-paste from i3 where it's been used and tested for a while now. So I still think it's an approach worth pursuing. :-) 

I just need to find some time to review this. Of course anyone else wanting to also review it is welcome to do so.

bebehei:
> Of course anyone else wanting to also review it is welcome to do so.

I must admit, that I'm following this PR since its start. I preached to @Stunkymonkey to write tests exactly about this literal `%` sign. Since this is a very stupid bug and easy to forget about in the next PR.

As he's short on time, I've written some for him. Another outcome of this was #289, which makes the testsuite effectively useable. @Airblader I would appreciate, if you take a look at #289 first.

bebehei:
Random find in the [`ddate` manpage](https://i3wm.org/i3status/manpage.html#_ddate):

> Note: Neither %. nor %X are implemented yet.

At least `%` is supported after this PR got merged ;-)

bebehei:
I found another corner case:

If a module has got the formatter `%percentage`, how do you print out the string `%percentage`? Is there a way to escape percents?

Airblader:
@bebehei Yeah that's true. I guess we could add some escaping syntax (though I would assume this is unlikely to happen). But let's definitely move that to a separate PR (we'd also have to fix it in i3)

Airblader:
@Stunkymonkey Can you drop a quick comment when this is ready? Also it needs a minor rebase now (sorry). Thanks for your work!

Stunkymonkey:
the format placeholder itself is done. But i wanted to look at the `START_COLOR` and `END_COLOR` functions.
maybe we could get rid of the outwalk variable.

Stunkymonkey:
@Airblader ready for review

src/format_placeholders.c
```diff
@@ -0,0 +1,71 @@
+/*
+ * vim:ts=4:sw=4:expandtab
+ *
+ * i3 - an improved dynamic tiling window manager
+ * © 2009 Michael Stapelberg and contributors (see also: LICENSE)
+ *
+ */
+// copied from  i3:libi3/format_placeholders.c and modified a bit
```
Airblader: What has been modified? I'd really like to avoid having this be any different between i3 and i3status.

src/print_cpu_temperature.c
```diff
@@ -1,37 +1,39 @@
...
+#include <limits.h>
 #include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
 #include <yajl/yajl_gen.h>
 #include <yajl/yajl_version.h>
 
 #include "i3status.h"
 
+#define STRING_SIZE 20
```
Airblader: Let's please not use these fixed-size strings. This goes for the other modules as well.

src/print_cpu_temperature.c
```diff
@@ -1,37 +1,39 @@
 // vim:ts=4:sw=4:expandtab
-#include <stdlib.h>
-#include <limits.h>
```
Airblader: Could we please remove all of this include noise? It makes this PR much bugger than it has to be without adding any benefit. We can do reformattings in separate PRs if there's a reason to do them.

src/print_cpu_usage.c
```diff
@@ -141,37 +147,34 @@ void print_cpu_usage(yajl_gen json_gen, char *buffer, const char *format, const
...
-                step *= 10;
-                padding++;
-            }
-            walk += strlen("cpu") + padding;
-        }
+    for (int number = 0; number < cpu_count; number++) {
+        int cpu_diff_idle = curr_cpus[number].idle - prev_cpus[number].idle;
+        int cpu_diff_total = curr_cpus[number].total - prev_cpus[number].total;
+        int cpu_diff_usage = (cpu_diff_total ? (1000 * (cpu_diff_total - cpu_diff_idle) / cpu_diff_total + 5) / 10 : 0);
+        // replace with placeholder_t.name an placeholder_t.value
```
Airblader: ?

src/print_cpu_temperature.c
Stunkymonkey: what do you suggest instead? allocating them on the heap an free instead?

src/print_cpu_temperature.c
Stunkymonkey: with a newer clang-format version it does this automatically
i reverted the changes

Stunkymonkey: or one macro for all?

src/format_placeholders.c
Stunkymonkey: https://github.com/i3/i3/pull/3319

src/print_cpu_temperature.c
Stunkymonkey: or do you prefer one macro for all?

src/print_volume.c
```diff
@@ -29,24 +29,23 @@
...
+    char string_percent[STRING_SIZE];
+    char string_volume[STRING_SIZE];
+
+    sprintf(string_percent, "%s", pct_mark);
+    sprintf(string_volume, "%d%s", ivolume, pct_mark);
+
+    placeholder_t placeholders[] = {
+        {.name = "%%", .value = string_percent},
+        {.name = "%volume", .value = string_volume}};
+
```
Stunkymonkey: @Airblader the percentage thing here is very bad, because printing percentage is now supported
but if we would remove this, we would have breaking changes

Airblader:
@Stunkymonkey I'm not sure when I have time to review this (my apologies). I do still think this approach is better, however, in the interest of fixing the bug, could you file a separate PR just with the initial fix commit? Then we get that merged for now to fix the bug and consider this a worthwhile refactoring which can be reviewed at some point.

Stunkymonkey:
ok fine.
i opened #297

Stunkymonkey:
the unnecessary `includes` are now all removed from my commits.

Stunkymonkey:
I think merging this will not happen any time soon.
@Airblader would it help if i split this PR into seperate smaller PRs?

Airblader:
Splitting it and addressing the open comment would be a good way to go, yes :-)

Issue #283, opened 2018-05-28
Shown IP address belongs to the wrong network interface.


duskCoder:
Hi gents,

The following pull request added a way to show the IP address of a network interface when a label is associated to the IP: https://github.com/i3/i3status/pull/213

To achieve this, if we have a network interface named **eth0**, instead of looking for an interface which name is exactly **eth0**, we look for a network interface which name _begins with_ **eth0**.

This change leads to a problem in a setup with two different network interfaces that have the same prefix:

Example of non-working setup:
```iproute2
8: foobar: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    inet 10.10.10.1/24 scope global foobar
       valid_lft forever preferred_lft forever
9: foo: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    inet 20.20.20.1/24 scope global foo
       valid_lft forever preferred_lft forever
```

With the following configuration file, meant to display the IP address of the network interface named **foo**:
```libconfuse
order += "ethernet foo"

ethernet foo {
    format_up = "%ip"
}
```

Here's what I obtain:
```sh
$ i3status -c /tmp/i3status.conf | head -n 1
10.10.10.1
```
But this IP address belongs to the interface named **foobar**.

**EDIT**: I should add that the problem occurs only if **foobar** has a network interface ID lesser than  **foo**. In the above example, foobar has the ID 8 and foo the ID 9.

Thank you by advance,
Olivier

Airblader:
Yeah, that sounds very possible. PRs to fix this welcome :-)

duskCoder:
Hi @Airblader ,

Thank you for your answer.

I can think of two different solutions.
- follow what was recommended by the author of the original pull request, @dlrobertson and support the following format:
```libconfuse
ethernet test0:dummyiface {
    [...]
}
```
This change might be cleaner but would make the patched version possibly incompatible with a configuration from version 2.12.
- keep the current format for the configuration and look for network interfaces that either match the entire name or _begin with_ the name and are followed by a `:` sign.
example, if the interface name is `foo`.
  * `foo` would match
  * `foo:bar` would match
  * `foobar` would **not** match

What are your thoughts ?
Olivier

dlrobertson:
@duskCoder I think allowing `<iface>:<label>` would be best I think, but I agree that it could be incompatible with prior versions.

[...]

PR #286, opened 2018-06-04
Fix issue #283 (shown IP belonging to wrong network interface)

duskCoder:
This fixes the behavior described in issue #283.

The first commit simplifies a little bit the algorithm used to loop through the structures returned by the `getifaddrs()` function. It makes the actual fix introduced in the second commit much more straightforward.

Kind regards,
Olivier

### DIFF
--- a/src/print_ip_addr.c
+++ b/src/print_ip_addr.c
@@ -37,22 +37,25 @@ const char *get_ip_addr(const char *interface, int family) {
         return NULL;
 
     /* Skip until we are at the input family address of interface */
-    for (addrp = ifaddr;
-
-         (addrp != NULL &&
-          (strncmp(addrp->ifa_name, interface, interface_len) != 0 ||
-           addrp->ifa_addr == NULL ||
-           addrp->ifa_addr->sa_family != family));
-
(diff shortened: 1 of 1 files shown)


src/print_ip_addr.c
```diff
@@ -29,30 +48,35 @@ const char *get_ip_addr(const char *interface, int family) {
...
     getifaddrs(&ifaddr);
 
     if (ifaddr == NULL)
         return NULL;
 
     /* Skip until we are at the input family address of interface */
-    for (addrp = ifaddr;
+    for (addrp = ifaddr; addrp != NULL; addrp = addrp->ifa_next) {
+        /* Strip the label if present in the .ifa_name field. */
+        strip_optional_label(addrp->ifa_name);
```
Airblader: I don't really like messing with the `addrp` object. I'd prefer if we stored the stripped name as a variable within the loop (and for that the function should just allocate the memory and return the string, then we free it in the loop).

duskCoder:
Hi @Airblader, 

Thank you for your feedback. I've updated the PR according to your input.

Thanks,
Olivier

Airblader:
Thanks!

PR #289, opened 2018-06-19
Testsuite

bebehei:
I wanted to write some tests for #285
This changes a few things in the testsuite:

- Run all tests, no matter if the previous had been successful or not
- Exposes the success in the exit-code, so travis and other scripts can have a reliable test if the testsuite actually fails or not
- Implement `make test`, so that `i3status` gets rebuilt before test by default. This makes switching branches much easier and you can run the tests locally on your computer
- Check if `i3status` crashed and also expose it. Stupid errors, which occur after the tested output are caught now, too!


Airblader:
Thanks!

eplanet:
Good improvement on testing! Thanks!

PR #290, opened 2018-06-20
Fix potential issues & avoid unnecessary instructions 

duskCoder:
Hi,

This pull request fixes the following issues discovered by reading the code and/or by enabling more compilation warnings:

* potential memory leak in `slurp_all_batteries()`
* invalid handling of errors when using `glob()`
* use of undefined macro `__OpenBSD__` when building with another OS
* undefined behavior when using `globfree()` after `glob()` failure.
* argument of `die()` and `fprintf()` not matching the requested format

In addition to that, the `die()` function and the `memory_absolute()` function got simplified by removing superfluous instructions.

Kind regards,
Olivier

### DIFF
--- a/i3status.c
+++ b/i3status.c
@@ -106,7 +106,7 @@ static bool path_exists(const char *path) {
 
 static void *scalloc(size_t size) {
     void *result = calloc(size, 1);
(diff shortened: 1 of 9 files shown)


Airblader:
This is great, I appreciate it. Thank you!

PR #291, opened 2018-06-24
Stop requiring CAP_NET_ADMIN

duskCoder:
Hi,

This pull request removes the now-unnecessary `CAP_NET_ADMIN` capacity from the `i3status` binary.

Querying an Ethernet interface speed has been allowed as a non-root user without this capacity since Linux kernel ~~2.6.36~~ (EDIT 2.6.37).

Speaking of which, I strongly recommend to run the following command as root to drop the capability from an installed `i3status` binary since it can be harmful from a security standpoint and is no longer needed.

```bash
setcap -r "$(which i3status)"
``` 

Kind regards,
Olivier


Airblader:
Thanks!

Issue #278, opened 2018-04-11
Cannot print percentage sign


endless-love:
I've tried: `%`, `%%`, `\%` in the format but none of them seems to work.

I'm using i3status 2.11 (2017-01-21)

Airblader:
Reading the code I think the issue is that we don't handle standalone % signs, effectively ignoring them. We need to make sure to go through all codepaths that replace placeholders and fix it.

Note that in libi3 we have a format_placeholders. It might make sense to think about reusing this.

Stunkymonkey:
i will provide a fix this weekend.

PR #297, opened 2018-07-13
able to print percentage

Stunkymonkey:
its now possible to have percentage before and after a variable. except
for the date. But percentage with dates does not make much sense to me, so
i skipped it. fixing #278


Airblader:
Thank you! Sorry for the trouble and thanks for the work you put in!

Stunkymonkey:
Thanks too.
(maybe you want to close #278?)

PR #299, opened 2018-07-15
Detect interface type on DragonFly BSD

tuxillo:
Currently there is no interface type detection in DragonFly which, on the default config, shows "down" for both wireless and ethernet devices.

**BEFORE:**

    no IPv6 | 31.9 GiB | W: down | E: down | No batteryNo battery | 1.05 |  | 2018-07-15 21:50:10

**AFTER:**

    no IPv6 | 31.9 GiB | W: down | E: 192.168.1.39 (1000baseT) | No batteryNo battery | 1.25 |  | 2018-07-15 21:50:45

I've also tried this patch on OpenBSD 6.1 but only with a Ethernet interface, it would be needed to be tested with a WLAN interface as well but I don't have any:

    # ./i3status -c i3status.conf 
    i3status: trying to auto-detect output_format setting
    i3status: auto-detected "term"
    i3status: Memory status information is not supported on this system
    no IPv6 | 636.7 MiB | W: down | E: 192.168.4.163 (1000baseT) | No battery | 1.23 |  | 2018-07-15 21:52:49
    # uname -a
    OpenBSD openbsd6.sector.int 6.1 GENERIC.MP#20 amd64

### DIFF
--- a/src/first_network_device.c
+++ b/src/first_network_device.c
@@ -3,13 +3,17 @@
 #include <sys/stat.h>
 #include <stdlib.h>
 #include <ifaddrs.h>
-#ifdef __OpenBSD__
+#if defined(__OpenBSD__) || defined(__DragonFly__)
 #include <sys/types.h>
 #include <sys/sockio.h>
 #include <sys/ioctl.h>
 #include <net/if.h>
+#endif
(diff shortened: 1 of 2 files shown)


Airblader:
Thank you!

PR #301, opened 2018-08-01
Fix for issue 300: Correctly print usage for cpu 10

wmarvel:
In cpu_print_usage.c, the step factor used to figure out how many extra characters to skip over when parsing through '%cpuN' directive is 10.

With a < comparison and a cpu directive of "%cpu10", the comparison fails, nothing is added to padding, only one character gets skipped, and the 0 character from "%cpu10" ends up being copied to the output buffer. This occurs only for CPU 10.

This PR changes that comparison to <= and adds a test for the issue. I've confirmed the test succeeds on hosts with more than, and less than, 10 cpus. With less than 10 cpus the test is set up to always succeed, in line with what the other cpu usage tests appear to be doing around having one or more than one cpu.

### DIFF
--- a/src/print_cpu_usage.c
+++ b/src/print_cpu_usage.c
@@ -163,7 +163,7 @@ void print_cpu_usage(yajl_gen json_gen, char *buffer, const char *format, const 
             }
             int padding = 1;
             int step = 10;
-            while (step < number) {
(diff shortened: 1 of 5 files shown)


Airblader:
Yes, thanks for the PR and also for the review!

PR #305, opened 2018-09-03
 Add support for "POWER_SUPPLY_STATUS=Not charging" (fixes #304) 

### LINKED ISSUE #304: Unknown battery status "Not charging"
Today I suddenly got a status_unk for my battery while my battery was full and the power supply still connected.
Looking at /sys/class/power_supply/BAT0/uevent there was a »POWER_SUPPLY_STATUS=Not charging«
That doesn't seem to be expected by i3status.
Now I'm not sure whether this status only happens when the battery is full or can happen at any time - but I guess it'd be best to handle it like POWER_SUPPLY_STATUS=Discharging?
The relevant source code seems to be line 182 in print_battery_info.c
If you think so, too, I'm happy to create a pull request just adding a »|| BEGINS_WITH(last, "POWER_SUPPLY_STATUS=Not charging")« to that line.

I'm using a Lenovo ThinkPad E540.
From BAT0/uevent:
```
POWER_SUPPLY_MODEL_NAME=LNV-45N1
POWER_SUPPLY_MANUFACTURER=LGC
POWER_SUPPLY_SERIAL_NUMBER=27831
```

David96:
As per the suggestion (see issue #304), here the pull request including a testcase.
"Not charging" is now handled the same way "Discharging" is.


Airblader:
Agreed. Thanks!

PR #312, opened 2018-10-03
strncpy + strlen is pointless

orestisf1993:


TwentyFourD:
I've just come across a similar occurrence (but with strncmp) in i3status.h:

#define BEGINS_WITH(haystack, needle) (strncmp(haystack, needle, strlen(needle)) == 0)

Does this fall into the same category?

orestisf1993:
No, that's correct. Compare:
```c
char *a = "Hello World!";
char *b = "Hello";
printf("%d %d\n", strcmp(a, b), strncmp(a, b, strlen(b)));
```

TwentyFourD:
Ah - yes. I was actually experimenting with:

bool begins_with(const char *haystack, const char *needle) {
    while (*needle)
        if (*needle++ != *haystack++)
            return false;
    return true;
}

which works, I think.

PR #315, opened 2018-10-18
man: fix typo for max_threshold in load

gportay:
Floating values use a dot (not a comma).

Fixes:

	invalid floating point value for option 'max_threshold'


orestisf1993:
Thanks!

PR #316, opened 2018-10-18
Switch to autotools

stapelberg:
Largely cargo-culted from i3lock and i3.

I might have missed something, though :).

Not tested on non-linux yet.
Maybe @jasperla could test on OpenBSD?
Maybe @bapt could test on FreeBSD?
Maybe @xtaran could test on kFreeBSD?
Maybe @sctincman could test on DragonflyBSD?
Maybe @afh could test on Mac?


stapelberg:
I’ll fix the travis setup tomorrow

configure.ac
```diff
@@ -0,0 +1,154 @@
...
+AC_DEFINE_UNQUOTED([I3STATUS_VERSION], ["${I3STATUS_VERSION}"], [i3status version])
+
+AX_CODE_COVERAGE
+
+dnl is_release must be lowercase because AX_CHECK_ENABLE_DEBUG calls m4_tolower
+dnl on its fourth argument.
+AX_CHECK_ENABLE_DEBUG([yes], , [UNUSED_NDEBUG], [$is_release])
+
+AC_PROG_CC_C99
+
```
bapt: You need:
PKG_PROG_PKG_CONFIG

otherwise pkg-config checks used later will fail (at least on freebsd)

stapelberg: Thanks, this is because our first check was in a conditional, see also https://autotools.io/pkgconfig/pkg_check_modules.html. I’ve restructured the checks, so things should work now. Can you confirm please?

stapelberg:
We already require GNU make in i3, and back then, I couldn’t find a way to express what I needed without GNU make.

If you want to, you’re welcome to submit PRs to i3, i3lock and i3status to make them independent of GNU make. Let’s treat that as a separate effort, though, and introduce autotools to i3status with GNU make for the time being :)

configure.ac
```diff
@@ -0,0 +1,154 @@
...
+	default_sanitizers=
+else
+	default_sanitizers=address
+fi
+AX_SANITIZERS(, [$default_sanitizers], [AC_DEFINE([I3STATUS_ASAN_ENABLED], [], [Enable ASAN])])
+
+AC_OUTPUT
+
+in_git_worktree=`git rev-parse --is-inside-work-tree 2>/dev/null`
+if [[ "$in_git_worktree" = "true" ]]; then
```
bapt: please use single [ and not doubled one, while freebsd for instance accept it, it is not supposed to be portable and might cause portability issues elsewhere

stapelberg:
Thanks for testing!

line 1888 in the Makefile you provided indeed looks incomplete:
```
am--fnord am--fnord-all $(TEST_LOGS) $(TEST_LOGS:
```

In my Makefile, the corresponding line (AFAICT) looks like this:
```
# Leading 'am--fnord' is there to ensure the list of targets does not
# expand to empty, as could happen e.g. with make check TESTS=''.
am--fnord $(TEST_LOGS) $(TEST_LOGS:.log=.trs): $(am__force_recheck)
am--force-recheck:
        @:
```

Maybe this is a kFreeBSD-specific automake issue?

configure.ac
stapelberg: Done.

stapelberg:
I don’t think we’ll get any more feedback here, so I’ll go ahead and merge the PR.

badboy:
I face the same problem as @xtaran. The build fails with `unterminated variable reference`. The corresponding Makefile line looks the same as above.
Running on Arch Linux.
Therefore I can't update the i3status-git package.

Additionally, the new way to build this is not documented in the README yet.

stapelberg:
Is this issue reproducible in an Arch Linux docker container? If so, can you provide the required steps please?

badboy:
Here's a quick one: https://github.com/badboy/arch-i3status

stapelberg:
I think I figured out the immediate cause: `make` is being run in the top directory (i3status) instead of the build directory (i3status/x86_64-pc-linux-gnu in the case of arch).

Changing the `make` line in the `PKGBUILD` file to `cd x86_64-pc-linux-gnu && make` makes compilation work.

I’m not familiar with `makepkg`, so I can’t say if there is a nicely integrated way to do builds in a separate build dir.

Also, perhaps we’re doing something weird which prevents `make` in the top directory from changing to the build directory automatically.

badboy:
Thanks! I think that's ok as a fix for now, but worth looking into a fix making a top-level `make` work

Airblader:
For the record, the i3status-git package does not suffer from this issue – we build in a build folder there.

badboy:
Oh, you already took care of that. Great! thank you.

PR #319, opened 2018-11-07
Read /proc/mounts if /etc/mtab can't be read

### LINKED ISSUE #257: i3status fails to run on WSL w/o /etc/mtab
Since WSL does not currently populate the /etc/mtab file, you must add a dummy file to boot a recent i3status process. This was tested with `i3status 2.11 (2017-01-21) © 2008 Michael Stapelberg and contributors` on Windows 10 (17063).

Running it with strace quickly spotted an ENOENT error right before the segfault. Adding an empty mtab seems to avoid the segfault for now but it might be better to handle cases where we're unable to read this file for one reason or another.

eplanet:
This would fix #257

### DIFF
--- a/src/print_disk_info.c
+++ b/src/print_disk_info.c
@@ -145,15 +145,22 @@ void print_disk_info(yajl_gen json_gen, char *buffer, const char *path, const ch
         if (strlen(sanitized) > 1 && sanitized[strlen(sanitized) - 1] == '/')
             sanitized[strlen(sanitized) - 1] = '\0';
         FILE *mntentfile = setmntent("/etc/mtab", "r");
-        struct mntent *m;
-
-        while ((m = getmntent(mntentfile)) != NULL) {
-            if (strcmp(m->mnt_dir, sanitized) == 0) {
-                mounted = true;
-                break;
-            }
+        if (mntentfile == NULL) {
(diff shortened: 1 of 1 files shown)


src/print_disk_info.c
```diff
@@ -145,15 +145,19 @@ void print_disk_info(yajl_gen json_gen, char *buffer, const char *path, const ch
         if (strlen(sanitized) > 1 && sanitized[strlen(sanitized) - 1] == '/')
             sanitized[strlen(sanitized) - 1] = '\0';
         FILE *mntentfile = setmntent("/etc/mtab", "r");
-        struct mntent *m;
+        if (mntentfile == NULL) {
```
orestisf1993:
```suggestion
        if (mntentfile == NULL) {
            mntentfile = setmntent("/proc/mount", "r");
        }
        if (mntentfile == NULL) {
```

eplanet: Well I wasn't sure file structure would be identically parsed between the two files. But if so, let's do that!

Thanks for the suggestion, but since there's a typo in the filename `/proc/mount` that should be `/proc/mounts`, I'm just adding it in the original commit if you don't mind.

orestisf1993:
No problem, we would have squashed them anyway

orestisf1993:
Thanks!

PR #320, opened 2018-11-07
Fix error messages

eplanet:
I noticed both these error messages weren't prefixed by `i3status: `. I moved CPU number parsing error to two distinct messages for clarity's sake.


Airblader:
Thanks!

Issue #303, opened 2018-08-12
Home timezone clock (feature request)


ArcturusB:
When moving to a different timezone, it’s convenient to have a clock of the home timezone that displays automatically. (Eg. the android deskclock does this.) Currently, this can only be done by manually by commenting and uncommenting a tztime module. 

This could be implemented by adding a boolean switch to the tztime module options. If the switch is on, only display the module when the locale timezone is different from the module’s timezone. 

I can submit a PR if you like this idea. :)

Airblader:
Hi,

thanks for raising this feature request. I do think this is simple & useful enough to include in i3status. Feel free to send a PR!

eplanet:
The related PR has been merged. That issue can be closed.

ArcturusB:
Many thanks for working on this @eplanet! I’m closing the issue.

PR #321, opened 2018-11-10
Add timezone switch

eplanet:
This adds the feature requested by #303.

The only drawback I see, is that when nothing should be displayed because the timezone is the same, then a [separator is anyway added](https://github.com/i3/i3status/blob/master/i3status.c#L681-L682).

I think we could fix that in a separate PR.

### DIFF
--- a/i3status.c
+++ b/i3status.c
@@ -399,6 +399,7 @@ int main(int argc, char *argv[]) {
         CFG_STR("timezone", "", CFGF_NONE),
         CFG_STR("locale", "", CFGF_NONE),
         CFG_STR("format_time", NULL, CFGF_NONE),
+        CFG_BOOL("only_when_tz_different", false, CFGF_NONE),
         CFG_CUSTOM_ALIGN_OPT,
         CFG_CUSTOM_MIN_WIDTH_OPT,
         CFG_CUSTOM_SEPARATOR_OPT,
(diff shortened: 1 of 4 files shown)


src/print_time.c
```diff
@@ -34,18 +34,28 @@ void set_timezone(const char *tz) {
...
+    set_timezone(NULL);
+    localtime_r(&t, &local_tm);
+
     set_timezone(tz);
     localtime_r(&t, &tm);
 
+    // When only_when_tz_different is true, compare local and target time to display only if different
+    time_t local_t = mktime(&local_tm);
+    double diff = difftime(local_t, t);
+    if (only_when_tz_different && diff == 0.0) {
```
Airblader: This does something other than advertised: what we want is to hide the time if the timezone is the same as the current system timezone. This, however, compares the times – but different timezones don't necessarily have different times.

i3status.c
```diff
@@ -399,6 +399,7 @@ int main(int argc, char *argv[]) {
         CFG_STR("timezone", "", CFGF_NONE),
         CFG_STR("locale", "", CFGF_NONE),
         CFG_STR("format_time", NULL, CFGF_NONE),
+        CFG_BOOL("only_when_tz_different", false, CFGF_NONE),
```
Airblader: I'd prefer naming this differently, maybe `hide_if_same_timezone`, for a couple reasons:

1. We should spell out "timezone". Not everyone might know the common abbreviation, and the timezone option itself is also spelled out.
2. "only when" doesn't contain a verb, it doesn't really say only *what*, so adding "hide" makes it a bit more clear.

eplanet:
Thanks for your review. I think it's better to resolve the separator issue in a distinct PR so that we are sure the issue is fixed independently of any feature. I hope I can work on that very soon!

Airblader:
Yes, a separate PR is better for that. We'll keep this one open in the meantime. :-)

eplanet:
I actually updated the PR to:
- Change boolean name to `hide_if_equals_localtime`: I think it's what it does and essentially people want to see their home time only if it's different, not if only the timezone name is different
- Fix the output to nothing when there's nothing to display. Actually `i3bar` handles the empty configuration block nicely already, so it's not necessary to fix it prior to merging that PR

PR #325, opened 2018-12-26
Implement %devicename specifier for volume module

Denton-L:
This commit implements the %devicename specifier for the volume module
for both PulseAudio and ALSA. This way, i3status will be able to display
the specific device that corresponds to the volume indicator.

Note that this is not implemented for the OSS API but is left in a state
where someone can pick it up for the future.

---

I implemented this because I have a pair of Bluetooth headphones that may not be connected. I'd like to check where the default sink is set to before blaring music so that I don't potentially annoy people in the same room.

Also, I tried to implement the OSS API portion but after mucking through the header files and the docuentation, I didn't want to implement something that may have had incorrect behaviour since I don't have a BSD system to test on.

### DIFF
--- a/src/print_volume.c
+++ b/src/print_volume.c
@@ -47,7 +47,7 @@
         fmt = fmt_muted;                                                                 \
     }
 
-static char *apply_volume_format(const char *fmt, char *outwalk, int ivolume) {
+static char *apply_volume_format(const char *fmt, char *outwalk, int ivolume, const char *devicename) {
     const char *walk = fmt;
 
     for (; *walk != '\0'; walk++) {
@@ -61,6 +61,10 @@ static char *apply_volume_format(const char *fmt, char *outwalk, int ivolume) {
         } else if (BEGINS_WITH(walk + 1, "volume")) {
             outwalk += sprintf(outwalk, "%d%s", ivolume, pct_mark);
             walk += strlen("volume");
+
+        } else if (BEGINS_WITH(walk + 1, "devicename")) {
+            outwalk += sprintf(outwalk, "%s", devicename);
+            walk += strlen("devicename");
 
         } else {
             *(outwalk++) = '%';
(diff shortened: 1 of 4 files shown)


Denton-L:
@Airblader if you get the chance, could you please review this?

I've been running it for just under a month now with no problems.

Airblader:
Hi! Sorry for the delay, I do still have this on my todo list, but haven't gotten around to it yet (and don't understand too much about the topic, hence it's not just a quick reivew for me :-) ).

@stapelberg If you have any spare time to review this, I'd appreciate it. Otherwise whenever I get around to it.

src/pulse.c
```diff
@@ -226,16 +248,49 @@ int volume_pulseaudio(uint32_t sink_idx, const char *sink_name) {
...
     pa_threaded_mainloop_lock(main_loop);
     get_sink_info(context, sink_idx, sink_name);
     pa_threaded_mainloop_unlock(main_loop);
     /* show 0 while we don't have this information */
     return 0;
 }
 
+bool description_pulseaudio(uint32_t sink_idx, const char *sink_name, char buffer[MAX_SINK_DESCRIPTION_LEN]) {
+    if (!context_ready || default_sink_idx == DEFAULT_SINK_INDEX)
+        return false;
```
stapelberg: nit: please use `{}` everywhere, even if the conditional body has only one line.

src/pulse.c
```diff
@@ -226,16 +248,49 @@ int volume_pulseaudio(uint32_t sink_idx, const char *sink_name) {
...
+     * when the description or volume actually changes, but we need it to be
+     * correct even if it never changes */
     pa_threaded_mainloop_lock(main_loop);
     get_sink_info(context, sink_idx, sink_name);
     pa_threaded_mainloop_unlock(main_loop);
     /* show 0 while we don't have this information */
     return 0;
 }
 
+bool description_pulseaudio(uint32_t sink_idx, const char *sink_name, char buffer[MAX_SINK_DESCRIPTION_LEN]) {
```
stapelberg: This can be made `static`, no?

src/print_volume.c
```diff
@@ -93,36 +97,51 @@ void print_volume(yajl_gen json_gen, char *buffer, const char *fmt, const char *
                                         !isdigit(device[strlen("pulse:")])
                                     ? device + strlen("pulse:")
                                     : NULL;
-        int cvolume = pulse_initialize() ? volume_pulseaudio(sink_idx, sink_name) : 0;
+        int cvolume = 0;
+        char description[MAX_SINK_DESCRIPTION_LEN];
```
stapelberg: Might be prudent to initialize `description` with `{'\0'}`

stapelberg:
Thanks!

PR #326, opened 2018-12-26
Reformat code using clang-format

Denton-L:
According to the contribution guidelines, code should be formatted with
clang-format before submitting a pull request. This change cleans up
code that previously did not conform so that future changes
pull-requests running clang-format will not affect unrelated files.


Denton-L:
My mistake, Arch Linux ships with clang-format 7.0.1 but I guess Travis uses 3.8 so there were some differences.

PR #327, opened 2018-12-26
ipv6: add support for %iface

Dettorer:
This follows the discussions in #205.

One problem is that in order to determine the IPv6's interface, we need to use `getifaddrs(3)` and walk through its linked list. We could return a cached value when we query the same IPv6 address twice in a row but that's not perfect, what do you think?

### DIFF
--- a/src/print_ipv6_addr.c
+++ b/src/print_ipv6_addr.c
@@ -4,6 +4,8 @@
 #include <sys/socket.h>
 #include <netinet/in.h>
 #include <arpa/inet.h>
+#include <net/if.h>
+#include <ifaddrs.h>
 
 #include <stdio.h>
 #include <stdbool.h>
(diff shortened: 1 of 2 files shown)


stapelberg:
Sorry for the lack of feedback on this. Are you still interested in pursuing this PR?

Dettorer:
Haha I saw the ping on the corresponding issue too. I don't personally use i3status anymore but I'll be happy to help finish this! Do you already have some feedback or do you want me to try and rebase it first? (I didn't realize it's been three whole years already)

stapelberg:
> Haha I saw the ping on the corresponding issue too. I don't personally use i3status anymore but I'll be happy to help finish this! Do you already have some feedback or do you want me to try and rebase it first? (I didn't realize it's been three whole years already)

Thanks!

The overall approach (using getifaddrs) is fine, I think.

Can you rebase the commit to the current git repository contents please?

Also, can you only call the interface-finding code when actually replacing `%iface`, so that users who don’t use `%iface` don’t need to incur the extra syscall cost?

Dettorer:
Hi! Sorry for the *year* of silence, I rebased and applied your suggestion, although with the new `format_placeholders()` function I had to do an extra scan of the `format_up` string with `strstr()` to know if getting the interface name was needed.

I tested the module with and without using the placeholder, it works and the function is called only when needed, but I didn't compare the execution time to know if using `strstr()` is actually faster than the function that finds the interface name (I guess it should be, and by a lot, as long as the `format_up` string isn't absurdly large).

Let me know if this needs more tweaking or testing, it shouldn't take me a year to react this time!

src/print_ipv6_addr.c
```diff
@@ -117,8 +119,65 @@ static char *get_ipv6_addr(void) {
...
+                break;
+            }
+        }
+    }
+
+    freeifaddrs(addresses);
+
+    if (*iface_string == '\0') {
+        /* We did not find the interface of the given address. */
+        return NULL;
```
stapelberg, line 168: Returning NULL here crashes format_placeholders(), so it would be better to just unconditionally return iface_string

src/print_ipv6_addr.c
```diff
@@ -117,8 +119,65 @@ static char *get_ipv6_addr(void) {
...
+    /* getifaddrs(3) returns a linked list of all the available IP addresses on
+     * the system. */
+    if (getifaddrs(&addresses) == -1) {
+        perror("getifaddrs()");
+        return NULL;
+    }
+
+    /* Search the linked list for our IP address */
+    for (struct ifaddrs *addr = addresses; addr != NULL; addr = addr->ifa_next) {
+        if (addr->ifa_addr && addr->ifa_addr->sa_family == AF_INET6) {
```
stapelberg, line 143: Invert this condition to save one level of indentation:

```c
if (addr->ifa_addr == NULL || addr->ifa_addr->sa_family != AF_INET6) {
  continue;
}
// …
```

src/print_ipv6_addr.c
```diff
@@ -117,8 +119,65 @@ static char *get_ipv6_addr(void) {
...
+/*
+ * Returns the name of the interface with which the given IPv6 address is
+ * associated.
+ * The return value is a statically allocated char* and mustn't be freed.
+ */
+char *get_iface_addr(const char *searched_addr_string) {
+    static char iface_string[IFNAMSIZ] = "";
+    struct ifaddrs *addresses;
+
+    if (searched_addr_string == NULL)
```
stapelberg, line 131: Add braces please (our style for new code always uses braces)

stapelberg (changes requested):

src/print_ipv6_addr.c
Dettorer: Indeed. Can I suggest something more explicit than an empty string, like "no iface found" or "missing_iface"? That would mean setting a default value for `iface_string` and returning it unconditionally. I guess I should also do that on library functions error rather than returning NULL, or return yet another string, like "error_iface" to help future debug?

stapelberg: Maybe returning `(error)` would be short enough, delimited enough (with the parenthesis), and descriptive enough?

Dettorer:
I applied the requested changes and the improved error reporting (replace the placeholder by `"(error)"` if any problem arises and print the details on stderr). I also made it so `get_iface_addr()` isn't called if the interface name placeholder is present but there is no IPv6 connectivity.

Dettorer:
I see the CI pipeline is failing because of a missing clang-format package, I ran the steps manually in the meantime.

A format issue was detected (fixed in 054833e from last force-push) and two tests don't pass (022-cpu-usage-tenth-cpu and 006-cpu-usage-max-threshold-format) but they fail on main too, it doesn't seem this is linked to this PR. Everything else (build and other tests) seems to run correctly.

stapelberg:
I pushed commit https://github.com/i3/i3status/commit/07ad5aef2deea1f10eb6c544593e7a87205f24f8, can you rebase your branch please?

stapelberg:
The tests all work on my machine (and on GitHub Actions: https://github.com/i3/i3status/actions/runs/3861330716/jobs/6582243202), both on main and with your changes.

Dettorer:
Done, and the CI does pass indeed (triggered on my fork by fast forwarding main to my branch https://github.com/Dettorer/i3status/actions/runs/3861423000)

stapelberg:
Thanks!

PR #332, opened 2019-01-23
Fix non Linux build for #309

eplanet:
This wasn't wrapped correctly in #309 and broke build on non Linux architecture (ex: MacOS).


orestisf1993:
Thanks!

PR #335, opened 2019-01-29
Support libcs without GLOB_TILDE

atlaua:
The GLOB_TILDE option to the glob() function is a GNU extension that's not supported by all libcs; for example, the musl libc doesn't have it.


Airblader:
Hi, thanks for the PR. However, this new implementation is not functionally equivalent as it won't properly expand things like `~` or `~user`.

atlaua:
I did indeed not consider the case ``path == "~"``; I've fixed that now. Thanks for the catch.

The lack of support for the ~user syntax, on the other hand, is intentional. The only user-supplied paths that are passed to resolve_tilde() are $XDG_CONFIG_HOME and $XDG_CONFIG_DIRS, and I don't see much use for the ability to put the path of another user's home directory into those variables.

However, I'd be happy to support this if you consider it useful; it'd only take a few extra lines of code after all.

PR #343, opened 2019-04-17
Increase buffer size for print_cpu_usage

tbodt:
My computer has 72 cores listed in /proc/stat, and after a few weeks of uptime, the numbers get big enough that it doesn't fit in 4096 bytes.


cliffspradlin:
I work with a computer whose /proc/stat is 12417 bytes long. I'd maybe suggest bumping up the buffer to 64K which ought to be enough for anyone!

stapelberg:
I implemented an alternative approach in commit https://github.com/i3/i3status/commit/e84f9588dfca3023f77fcc7d320c70945d852f9d: switching to the buffered file API, so that we are out of the sizing business (except for individual line size, for which 4096 should be plenty).

PR #344, opened 2019-04-23
print_cpu_usage: double buffer size

stapelberg:
On systems with many cores, the current 4096 bytes are not sufficient to hold the contents of the /proc file.

### DIFF
--- a/src/print_cpu_usage.c
+++ b/src/print_cpu_usage.c
@@ -76,7 +76,7 @@ void print_cpu_usage(yajl_gen json_gen, char *buffer, const char *format, const 
     }
 
     memcpy(curr_cpus, prev_cpus, cpu_count * sizeof(struct cpu_usage));
-    char buf[4096];
+    char buf[8192];
     curr_cpu_count = get_nprocs();
     if (!slurp(path, buf, sizeof(buf)))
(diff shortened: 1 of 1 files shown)


Airblader:
#343?

Airblader:
BTW, the reason I didn't merge the other one yet is simply because I was thinking that instead of creating large buffers it might be better to just read the filesize first. Haven't had the time to worry about it yet, but I assume this means you're fine with just making the buffer bigger ;-)

stapelberg:
Ah, thanks, I didn’t check existing issues before. Given the comments on #343, I think it indeed makes sense to read the file size using fstat after opening the file.

orestisf1993:
You can't read the file size of `/proc/stat`: https://unix.stackexchange.com/questions/265232/why-would-the-size-of-proc-stat-show-as-zero

We'll need to dynamically allocate memory

Airblader:
But does that also apply when using fstat after opening the file as @stapelberg mentioned? I would think it should work then.

orestisf1993:
Yes, it still won't work AFAICT

tbodt:
Most of the files in /proc, including /proc/stat, have no size.

PR #331, opened 2019-01-17
Added feature to print contents from file

tuggan:
I've added a function to print contents from a file to the statusbar (with newlines removed).
---
I implemented this to print information from /sys/ and /proc/ in my statusbar but it can print information from any readable file.

PR #347, opened 2019-06-10
print_file_contents: allocate enough to include \0

thrasymache:
related to #331
Previously, if max_chars was read, the null byte would be written
past the end of buf.


Airblader:
Thanks!

PR #356, opened 2019-07-09
add 'Text' module

elbeardmorez:
-facilitates bespoke literal string blocks of arbitrary colour

..and as per the man page entry:

Intended for use as string literals and placeholders for post-processing.

...

Hardly an Earth shattering addition, but this eased my requirements whilst setting up the net speed (contrib) hack. Primary, I refused to abuse / pollute other blocks to get the job done (e.g. placeholders in other format lines ..which adds unwanted chaining). For literals, I wanted arbitrary colours without the indirect (and limiting) configuration of a named colour.

Thanks for the tool.

pic:
2019Jul09 i3status text

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -68,6 +68,7 @@ i3status_SOURCES = \
 	src/print_volume.c \
 	src/print_wireless_info.c \
 	src/print_file_contents.c \
+	src/print_text.c \
 	src/process_runs.c \
 	src/pulse.c
 
--- a/i3status.c
+++ b/i3status.c
@@ -503,6 +503,15 @@ int main(int argc, char *argv[]) {
         CFG_CUSTOM_SEP_BLOCK_WIDTH_OPT,
         CFG_END()};
 
+    cfg_opt_t text_opts[] = {
+        CFG_STR("text", NULL, CFGF_NONE),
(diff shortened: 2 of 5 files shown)


elbeardmorez:
wrt. the build error:
```
Code was not formatted using clang-format!
```
..that's inaccurate. I read the contrib note, I used the clang formatting command, I then backed out what I saw was an incredibly dumb alignment change - you don't align parts of the if statement to the block, period.

I noted that running the command also changed other sections unrelated to my patch, hence I assumed that some leeway / manual intervention was desirable.

Airblader:
I think this can be addressed reasonably by using the module we added in #331 if you don't want to put the string into other blocks, thus I am leaning towards not merging this. @stapelberg Your thoughts? 

Regarding the formatting, likely you're using a different version of clang-format.

stapelberg:
> I think this can be addressed reasonably by using the module we added in #331 if you don't want to put the string into other blocks, thus I am leaning towards not merging this. @stapelberg Your thoughts?

Agreed.

Airblader:
Alright, then let's close this.

@elbeardmorez Thanks for the PR in any case, though. I'm sorry we're deciding to reject this; in general we'd always recommend to first open an issue to see if a PR would be expected or whether there's alternatives etc. Thanks for understanding!

PR #357, opened 2019-07-16
make pulseaudio optional

Piraty:
I'm no autohell guru, but this could be used as a starting point for simplification and modularity of `configure.ac`.


stapelberg:
Thanks for sending a PR, but we’re doing it this way intentionally. See also https://michael.stapelberg.ch/posts/2019-05-23-optional-dependencies/ for more thoughts on the topic.

PR #364, opened 2019-08-31
configure.ac: exclude build without running git

tharvik:
Currently, `configure.ac` runs `git rev-parse` inconditionally in order to remove the build dir automatically from git tracking. But, it may travel outside of the build directory, finding a gitdir away from the srcdir, for eg. when distributing an archive, thus modifing it. When using a sandbox to build it, it catches that but also disallow to run it safely.

This PR avoid reading the gitdir before it's checked for existence thus avoiding this issue.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -144,15 +144,12 @@ AX_SANITIZERS(, [$default_sanitizers], [AC_DEFINE([I3STATUS_ASAN_ENABLED], [], [
 
 AC_OUTPUT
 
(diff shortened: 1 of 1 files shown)


orestisfl:
See https://github.com/i3/i3/issues/2611 as well

tharvik:
It's been a year and I'm regularly having this issue when building it on my machine.
Is there anything I can do to help with getting it merge?

PR #366, opened 2019-09-19
Fix propagation of read error from slurp.

duskCoder:
Hi,

I came across the `slurp()` function "succeeding" even if the underlying call to `read()` fails. When that happens the destination buffer is left uninitialized and the caller has no way to figure it out.

Although it is quite unlikely that a single call to `read()` would ever fail right after a successful call to `open(..., O_RDONLY)`, it is possible, for instance, if the file being opened is a directory.

Fixed by propagating the error (i.e. returning `false` from `slurp()`) if the call to `read()` fails.

I've also updated the documentation of the function to emphasize that the destination buffer content should not be relied upon if the `slurp()` function fails.

Hope this helps,
Olivier

### DIFF
--- a/src/general.c
+++ b/src/general.c
@@ -27,7 +27,7 @@ bool slurp(const char *filename, char *destination, int size) {
         destination[n] = '\0';
     (void)close(fd);
 
-    return true;
+    return n != -1;
 }
 
 /*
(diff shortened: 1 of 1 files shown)


stapelberg:
Thank you!

PR #369, opened 2019-10-26
Introduce memory options 'unit' and 'decimals'

31KM:
Previously the format placeholders were auto-converted to the maximum possible
unit, e.g. /proc/meminfo reports MemTotal of 16307104kB which will get
converted to 15.6GiB. It is now possible to specifiy the target unit, e.g. Mi,
which will be used for the conversion - in the example it would lead to
15924.9MiB.

The resulting number can now be further formatted via the decimal option. It
allows to specify the number of decimals to use, e.g. 15.6GiB vs. 15GiB or
15924.9MiB vs. 15925MiB.


Airblader:
Makes sense to me and implementation looks simple & clean. Thanks!

PR #370, opened 2019-10-28
print_mem: Add memory monitor support on FreeBSD

Partmedia:
Support some memory stats on FreeBSD using sysctl(3). Not all monitors are available, so set those to zero.


centromere:
This patch works perfectly for me on 12.1-RELEASE. Thanks!

Partmedia:
Thanks for testing!

centromere:
Note that without setting `ram_available`, we are unable to take advantage of the `threshold_degraded` and `threshold_critical` features. Personally, I've set `ram_available = ram_free` for now.

Partmedia:
That's true, but that also might be a little misleading. FreeBSD tries to utilize all available system memory by using available (not being used by programs or the kernel) memory as disk cache. So when free memory is low, it may not be that applications don't have enough memory left to run without swapping, but that the disk cache has gotten big.

centromere:
Indeed. What do you think is a good way to estimate the available memory?

`vm.stats.vm.v_inactive_count` + `vm.stats.vm.v_cache_count` + `vm.stats.vm.v_free_count` perhaps?

PR #378, opened 2020-01-17
Add support for batteries outputting Time to empty on linux

C-Elegans:
The battery in the pinebook pro does not output the remaining charge
in Wh or mAh, so i3 cannot calculate the time remaining. However, it
does directly output the number of minutes remaining on in
POWER_SUPPLY_TIME_TO_EMPTY. This adds support for reading this field
and converting it to seconds_remaining, so i3status can report the time remaining on the battery


src/print_battery_info.c
```diff
@@ -173,6 +173,8 @@ static bool slurp_battery_info(struct battery_info *batt_info, yajl_gen json_gen
             batt_info->present_rate = abs(atoi(walk + 1));
         else if (BEGINS_WITH(last, "POWER_SUPPLY_VOLTAGE_NOW="))
             voltage = abs(atoi(walk + 1));
+        else if (BEGINS_WITH(last, "POWER_SUPPLY_TIME_TO_EMPTY_NOW="))
```
Airblader: Could you maybe add a testcase for this?

C-Elegans: Yup, added test 026-battery-time-to-empty

rillian:
What path do you have in your `i3status.conf`? I still get "No battery" with this patch. If I add `path = "/sys/class/power_supply/%d-battery/uevent"` to pick up the `cw2015-battery` device (`%d` is turned into a `*` glob) I get `BAT 0.00% 00:00` or  `CHR 0,00%`. That is, state is reflected but time remaining is not. Neither is percentage reported if I add a line parsing `POWER_SUPPLY_CAPACITY`.

This is with a local build of the tsys 5.5.0 kernel. Does your kernel report battery capacity differently?

C-Elegans:
@rillian I specify the full path to the battery, no substitution with %d:

```
 battery 0 {
        status_chr = " CHR"
        status_bat = " BAT"
        format = "%status %percentage %remaining %consumption"
        path = "/sys/class/power_supply/cw2015-battery/uevent"
        low_threshold = 20
        threshold_type = "percentage"
}
```

palfrey:
Running into the same issue on my Pinebook Pro. Is there anything in particular I can do to aid this getting merged?

AeroNotix:
damn I just forked to send this same PR! Oh well.

stapelberg:
Thank you!

PR #380, opened 2020-02-10
use format_placeholder(i3lib) for battery_info

Stunkymonkey:
its now possible to have percentage before and after a variable.
each module is doing the same procedure in replacing the `%variable` with a real value. I think it would be better to provide/use an function for this, because it will cause less errors in the future.

this PR is based on #285


Stunkymonkey:
i started with `battery_info` to keep the PR small.
this work can be used by all others `print`-modules

Airblader:
Thank you!

Issue #384, opened 2020-02-18
use format_placeholders in all print-modules


Stunkymonkey:
Just to keep track what is already implemented:
- [X] `battery_info`
- [x] `cpu_temperature`
- [ ] `cpu_usage`
- [x] `ddate`
- [x] `disk_info`
- [x] `eth_info`
- [x] `file_contents`
- [x] `ip_addr`
- [x] `ipv6_addr`
- [x] `load`
- [x] `mem`
- [x] `path_exists`
- [x] `run_watch`
- [x] `time`
- [x] `volume`
- [x] `wireless_info`

@Airblader is it wanted to use `format_placeholders` in all modules? even in `cpu_usage`?

Airblader:
In general yes, I'd try to do this consistently. If there's a good reason not to use it in one of the modules we should evaluate them.

Stunkymonkey:
@Airblader can you have a quick look at the `cpu_usage` code and think about how we are implementing this with format_placeholders?
What we definetly do not want to do; is allocating one giant placeholder struct; which then would have a constant size.

Airblader:
If there's a good reason for not using the function in that module then I'm also fine leaving that one out.

Stunkymonkey:
complexity wise i would like to leave out the `cpu_usage`.

If we would implement it, we would have to allocate one giant placeholder which contains all variables for all cores which. This means for a multi-core machine (e.g. 96cores) if would be a little memory overhead.
We would not only allocate the memory but also call the `snprintf` the same amount of times. Even if we would only use a single variable out of all of them.

Stunkymonkey:
@Airblader can we close this issue?

Stunkymonkey:
ping @orestisfl ?

PR #386, opened 2020-02-18
use format_placeholder for ddate

Stunkymonkey:
related to #384

### DIFF
--- a/src/print_ddate.c
+++ b/src/print_ddate.c
@@ -111,9 +111,9 @@ void print_ddate(yajl_gen json_gen, char *buffer, const char *format, time_t t) 
     char string_H[STRING_SIZE];
     char string_N[STRING_SIZE];
     /* Newline- and Tabbing-characters */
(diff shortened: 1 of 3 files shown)


src/print_ddate.c
```diff
@@ -202,17 +94,83 @@ struct disc_time *get_ddate(struct tm *current_tm) {
...
+    char string_a[STRING_SIZE];
+    char string_B[STRING_SIZE];
+    char string_b[STRING_SIZE];
+    char string_d[STRING_SIZE];
+    char string_e[STRING_SIZE];
+    char string_Y[STRING_SIZE];
+    char string_H[STRING_SIZE];
+    char string_N[STRING_SIZE];
+    /* Newline- and Tabbing-characters */
+    char string_n[STRING_SIZE] = "\n";
```
Airblader: For these we know the exact size, can't we use that?

Stunkymonkey: yeah sure. i will change this

PR #387, opened 2020-02-20
Small wrapper with the ability to add custom info in any position

ilya-bystrov:
The idea of this wrapper is to add custom info in any position.
I didn't find anything similar in `contrib` directory.

### DIFF
--- a/contrib/any_position_wrapper.sh
+++ b/contrib/any_position_wrapper.sh
@@ -1,0 +1,64 @@
+#!/usr/bin/env bash
+
+# This i3status wrapper allows to add custom information in any position of the statusline
+# It was developed for i3bar (JSON format)
+
+# The idea is to define "holder" modules in i3status config and then replace them
+
+# In order to make this example work you need to add
+# order += "tztime holder__hey_man"
+# and 
+# tztime holder__hey_man {
+#        format = "holder__hey_man"
+# }
+# in i3staus config 
+
+# Don't forget that i3status config should contain:
+# general {
+#   output_format = i3bar
+# }
+#
+# and i3 config should contain:
+# bar {
+#   status_command exec /path/to/this/script.sh
+# }
+
+# Make sure jq is installed
+# That's it
+
+# You can easily add multiple custom modules using additional "holders"
+
+function update_holder {
+
+  local instance="$1"
+  local replacement="$2"
+  echo "$json_array" | jq --argjson arg_j "$replacement" "(.[] | (select(.instance==\"$instance\"))) |= \$arg_j" 
+}
+
+function remove_holder {
+
+  local instance="$1"
+  echo "$json_array" | jq "del(.[] | (select(.instance==\"$instance\")))"
+}
+
+function hey_man {
+
+  local rand_val=$((RANDOM % 3))
+  if [ $rand_val == 1 ] ; then
+    local json='{ "full_text": "Hey Man!", "color": "#00FF00" }'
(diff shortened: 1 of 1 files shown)


contrib/wrapper.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+
+# This i3status wrapper allows to add custom information in any position of the bar
+# It was developed for i3bar (JSON format)
+
+# The idea is to define "holder" modules in i3status config and than replace them
```
Airblader: s/than/then/

contrib/wrapper.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# In order to make this example works you need to add
```
Airblader: s/works/work/

contrib/wrapper.sh
```diff
@@ -0,0 +1,48 @@
...
+# order += "tztime holder__hey_man"
+# and 
+# tztime holder__hey_man {
+#        format = "holder__hey_man"
+# }
+# in i3staus config 
+
+# Don't forget that
```
Airblader: Why is there a line break after just three words?

contrib/wrapper.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
```
Airblader: Can we give this a more specific name than `wrapper.sh`? Something that conveys what it does would be good.

I know two other scripts in here are already named wrapper, but they're seven years old and if anything it proves my point... :-)

contrib/wrapper.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# This i3status wrapper allows to add custom information in any position of the bar
```
Airblader: In any position of the *statusline*, not the bar (you cannot place something left of the workspaces with this).

contrib/wrapper.sh
```diff
@@ -0,0 +1,48 @@
...
+# line
+
+function update_holder {
+
+  local instance="$1"
+  local replacement="{ \"full_text\": \"$2\" }"
+  echo "$json_array" | jq --argjson arg_j "$replacement" "(.[] | (select(.instance==\"$instance\"))) |= \$arg_j" 
+}
+
+i3status | (read line; echo "$line"; read line ; echo "$line" ; read line ; echo "$line" ; while true
```
Airblader: Why are we read-echoing three times here? Shouldn't it just be twice (version + start of array)?

ilya-bystrov: Here it is. Have to wait for comma prefixed JSON-array.

```
$ i3status | while true; do read line ; echo $line ; echo '==='; done
{"version":1}
===
[
===
[{"name":"disk_info","instance":"/","markup":"none","full_text":"45.2 GiB"}]
===
,[{"name":"disk_info","instance":"/","markup":"none","full_text":"45.2 GiB"}]
===
,[{"name":"disk_info","instance":"/","markup":"none","full_text":"45.2 GiB"}]
===
```

contrib/wrapper.sh
ilya-bystrov: My intention was to be able to place custom module in any position.
So I suggest `any_position_wrapper.sh`

ilya-bystrov:
> Thanks! As a final step, could you please squash the commits into a single one?
Sure, I've done it

ilya-bystrov:
@Airblader up

Airblader:
Thank you! I apologize for the delay :-)

PR #391, opened 2020-03-11
print_volume: Native NetBSD audio API support

niacat:
Avoid using the OSS emulation layer to fetch the master volume. Share code with OpenBSD, who inherited this API from us.

Avoid a failed read on the master control's mute enum if it can't be found, because sometimes there isn't one.


Airblader:
Thanks!

PR #392, opened 2020-03-12
add autoconf to requirements

banderlog:


Airblader:
Can you maybe add in parentheses that it is a compile-time only dependency and then squash your commits? Thanks!

orestisfl:
Thanks!

PR #394, opened 2020-03-29
avoid out-of-bounds read after invalid %cpu conversion

smaeul:
In the case where no CPU number is given, skipping a character of
padding actually skips the null terminator, causing further iterations
through the loop to read out of bounds. Have sscanf() return the number
of characters read, instead of reconstructing it from the CPU number.

This was observed as a failure in test 024-cpu-usage-invalid-cpu.


Airblader:
Thank you for this!

PR #395, opened 2020-03-29
perform few minor improvements in i3status.man

myTerminal:
While re-doing my configuration for [i3](https://github.com/i3/i3) and [i3status](https://github.com/i3/i3status) into my [dotfiles](http://github.com/myTerminal/dotfiles) after almost five years since I moved to i3, I read the entire [man page for i3status](https://i3wm.org/i3status/manpage.html), found a few minor things that could probably be slightly improved and hence thought of contributing with the below two changes:

- Replace `red` and `yellow` with directive names `color_bad` and `color_degraded` to make it more specific
- Format directives for colors and all other ones I could to make them go from example `low_threshold` to `+low_threshold+`

Kindly let me know if something doesn't look right.

---
This change is [Reviewable](https://reviewable.io/reviews/i3/i3status/395)

### DIFF
--- a/man/i3status.man
+++ b/man/i3status.man
@@ -235,8 +235,8 @@ best available public IPv6 address on your computer).
 
 Gets used, free, available and total amount of bytes on the given mounted filesystem.
 
-These values can also be expressed in percentages with the percentage_used,
-percentage_free, percentage_avail and percentage_used_of_avail formats.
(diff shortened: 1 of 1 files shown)


Airblader:
Thanks so much for this, it is greatly appreciated!

PR #402, opened 2020-03-31
use format_placeholder for wireless_info

Stunkymonkey:
related to #384

### DIFF
--- a/src/print_wireless_info.c
+++ b/src/print_wireless_info.c
@@ -519,7 +519,12 @@ void print_wireless_info(yajl_gen json_gen, char *buffer, const char *interface,
         if (ipv6_address == NULL) {
             START_COLOR("color_bad");
             outwalk += sprintf(outwalk, "%s", format_down);
-            goto out;
+
+            END_COLOR;
+            free(ipv4_address);
+            free(ipv6_address);
+            OUTPUT_FULL_TEXT(buffer);
+            return;
         } else {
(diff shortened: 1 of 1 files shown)


Airblader:
@Stunkymonkey Can you look into the failed test? Thanks!

src/print_wireless_info.c
```diff
@@ -517,7 +519,12 @@ void print_wireless_info(yajl_gen json_gen, char *buffer, const char *interface,
         if (ipv6_address == NULL) {
             START_COLOR("color_bad");
             outwalk += sprintf(outwalk, "%s", format_down);
-            goto out;
```
Airblader, line 520: Why was this necessary? This now just duplicates what happens in `out` except for the placeholders, right?

Stunkymonkey: sure, but otherwise i could not distinguish in the end between "interface down/up". If the interface is up i have to free the buffer after the printing.

Stunkymonkey: and having `goto` labels is not always lucid

Airblader: > If the interface is up i have to free the buffer after the printing.

Right, point taken.

> and having goto labels is not always lucid

There are several ways to avoid duplicating code; macros, functions, goto. ;-) Since in this case it's actually (slightly) different and not too much I think it's OK.

PR #403, opened 2020-04-01
Update net-speed.sh

YaguraStation:
Added the common prefix `enx` as possible string for interface detection. See https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/


Airblader:
Thank you!

PR #405, opened 2020-04-04
use statvfs f_frsize for file system block size on NetBSD

tripledonkey:
First ever pull request, please be gentle!  I noticed disk usage was reporting wrong on NetBSD, I found this to be because the file system block size is actually in f_frsize of statvfs struct on NetBSD, I confirmed this with some of the NetBSD developers (nia and ryoshu).

### DIFF
--- a/src/print_disk_info.c
+++ b/src/print_disk_info.c
@@ -73,17 +73,9 @@ static bool below_threshold(struct statvfs buf, const char *prefix_type, const c
     } else if (strcasecmp(threshold_type, "percentage_avail") == 0) {
         return 100.0 * (double)buf.f_bavail / (double)buf.f_blocks < low_threshold;
     } else if (strcasecmp(threshold_type, "bytes_free") == 0) {
-#if defined(__NetBSD__)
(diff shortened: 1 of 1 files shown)


Airblader:
Thanks for the PR! And also for confirming with some NetBSD people.

@stapelberg Do you have an opinion on this? I can't speak for NetBSD and also don't much know these system APIs. Any chance for non-NetBSD we can do the same?

stapelberg:
TIL: https://stackoverflow.com/a/54835515/712014 (also https://debbugs.gnu.org/cgi/bugreport.cgi?bug=11406)

Yes, I think we should use f_frsize on all systems.

Airblader:
Thanks, Michael. @tripledonkey Can you change the PR accordingly? Thank you!

tripledonkey:
I don't really know my way around git.  I have now changed the code so that it not specific to NetBSD, but how do I amend the PR?  What git commands should I run?  Sorry for being clueless on this.

stapelberg:
git commit -a to amend the previous commit
git push -f to update the PR

stapelberg:
Thanks!

PR #409, opened 2020-04-26
Format placeholder volume

Stunkymonkey:
related to #384


Stunkymonkey:
this code is not very beautiful. But the `print_volume` file was already super messy.

PR #411, opened 2020-04-30
file_contents: Handle utf-8 characters

### LINKED ISSUE #410: Feature request: handle clipped UTF-8 in print_file_contents()
* I use a `read_file` block to print a file whose contents may contain non-ASCII characters encoded in UTF-8.
* I set `Max_characters` to 120.
* It looks like `print_file_contents()` doesn't attempt to observe logical boundaries of UTF-8 octets, so there are cases where the printed line ends in gibberish. (Side effect - this messes with my Pango markup.) For example, this sequence of three bytes followed by the four-byte symbol "🈚":
```
foo🈚🈚🈚🈚🈚...[snip]
```
would reproduce the issue for me.

My naive reading of RFC 3629 suggests that it's easy to detect UTF-8 octet sequences. *May I submit a pull request to further truncate file contents upon encountering clipped UTF-8?*

orestisfl:
Just testing the waters here, documentation needs to be updated

Fixes #410

@ j39m can you check this please?

### DIFF
--- a/include/i3status.h
+++ b/include/i3status.h
@@ -31,6 +31,7 @@ extern char *pct_mark;
 
 #define BEGINS_WITH(haystack, needle) (strncmp(haystack, needle, strlen(needle)) == 0)
 #define max(a, b) ((a) > (b) ? (a) : (b))
+#define min(x, y) ((x) < (y) ? (x) : (y))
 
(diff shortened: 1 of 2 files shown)


orestisfl:
@j39m can you check this please?

PR #426, opened 2020-07-27
Implement keyboard layout indicator

spektom:
Use Jay Bromley's XKeyboard class for displaying current keyboard layout.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -46,12 +46,15 @@ i3status_LDADD = \
 	$(PULSE_LIBS) \
 	$(NLGENL_LIBS) \
 	$(ALSA_LIBS) \
-	$(PTHREAD_LIBS)
+	$(PTHREAD_LIBS) \
+	-lX11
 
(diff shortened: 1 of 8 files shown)


Airblader:
See #279, #174 and #34. 

Pinging @stapelberg in case he wants to reconsider and reopen this (or one of the others).

stapelberg:
I don’t see a need to reopen keyboard layout indicators for now.

Most people use just one keyboard layout, few use two, even fewer use more. The limited user base means the cost of this feature (additional build- and runtime dependencies, customization concerns, etc.) doesn’t outweigh its usefulness.

When using more than one keyboard layout, just modify your switching script/keybinding/method to write a file with your preferred layout identifier and use i3status’s `read_file` functionality to include the file contents.

spektom:
I've written a small program that solves the issue, based on your suggestion: https://github.com/spektom/i3status-kbd-layout
Thanks!

PR #427, opened 2020-07-29
Fix crash in print_memory

Gravemind:
Fixes a stack-overflow when memory is, for example, "1020.17 TiB".

This fix limits the max number of memory decimals to 4.

The crash was probably introduced in 066e813331 .


stapelberg (approved):

stapelberg:
Thank you!

PR #429, opened 2020-08-12
Add keyboard layout status.

alborzjafari:
libxkbfile-dev and libx11-dev should be installed.

i3status.conf:
...
order += "keyboard_layout
...

keyboard_layout {
}
...


Airblader:
See #426 and others

PR #430, opened 2020-08-29
Allow selective pango-markup in tztime-module

### LINKED ISSUE #421: i3status does not allow selective pango-markup in tztime-module
`i3status` only supports global pango-markup in the tztime-module via the `format_time`-helper construct. This does not allow for selective markup, for example to aid readability by distinguishing clock and date by color.

Minimal (desired) config example:
```
general {
        colors = true
        interval = 60
        markup = "pango"
}

order += "tztime local"

tztime local {
        format = "%a <span foreground='#bbbbbb'>%Y-%m-%d</span> %H:%M"
}
```

Expected result:
`Thu 2020-06-11 22:00` with slightly darker `2020-06-11`.

Actual result:
`Thu &lt;span foreground=&apos;#bbbbbb&apos;&gt;2020-06-11&lt;/span&gt; 22:00`, all in one color.

As the `format_time`-directive is a single hardcoded option, this is not capable of producing the desired markup.

CC @sur5r

cherti:
This PR implements selective pango-markup in i3status (and therefore closes #421).

It makes two assumptions:
1. Results from time formatting will not introduce specifiers that will interfere with the pango-markup.
This is an assumption, but I would argue a justified one as I do not see which kind of character time formatting would (realisticly) introduce that could interfere here. One could certainly write a locale where weekday-names are in a form that this will happen, but I am not aware of any real-world locales that have such an issue.

2. The total string length will not exceed 511 characters.
This is an assumption based on the following observation:
Assume one would create a tzdata-element of the form `%Y-%m-%d %H:%M, %a`. This would be date as well as current time + the name day of the day of the week. Further assumed we would now want to give every single entry a different foreground and background color and font weight, i.e. surrounding it by a markup of type `<span background='#ffffff' foreground='#000000' weight='bold'>…</span>`. Then the resulting markup would cover 414 of 511 characters, still yielding 97 characters for the actual time strings (and possible further markup), which should therefore be a sufficiently conservative estimate.

Alternatively, instead of using a buffer size of 512, one could compute the length of the resulting string, by assuming an expansion length of qualifiers and then computing a buffer size based on the length of the actual format string, extended by this correction factor. However, this would introduce more code and be more CPU-intensive for possibly very limited gain, therefore I think the current approach seems to be preferrable.

CC @stapelberg for review, as mentioned in https://github.com/i3/i3status/issues/421

CC @sur5r

### DIFF
--- a/src/print_time.c
+++ b/src/print_time.c
@@ -64,8 +64,7 @@ void print_time(yajl_gen json_gen, char *buffer, const char *title, const char *
     char string_time[STRING_SIZE];
 
     if (format_time == NULL) {
-        strftime(string_time, sizeof(string_time), format, &tm);
-        maybe_escape_markup(string_time, &outwalk);
+        outwalk += strftime(buffer, 4096, format, &tm);
     } else {
         strftime(string_time, sizeof(string_time), format_time, &tm);
         placeholder_t placeholders[] = {


sur5r:
1. Agreed.
2. I'm not sure about the limit. `strftime` writes to the shared `buffer` which is 4K but AIUI has to fit all the plugins. So it's not so easy to tell how much space is left.

stapelberg:
> 2\. I'm not sure about the limit. `strftime` writes to the shared `buffer` which is 4K but AIUI has to fit all the plugins. So it's not so easy to tell how much space is left.

`buffer` is not shared between plugins. Each plugin can make full use of `buffer`. See https://github.com/i3/i3status/blob/a3a37da11e4b7aad293eda645761f342e5eafc3b/i3status.c#L641-L647

src/print_time.c
```diff
@@ -64,8 +64,7 @@ void print_time(yajl_gen json_gen, char *buffer, const char *title, const char *
     char string_time[STRING_SIZE];
 
     if (format_time == NULL) {
-        strftime(string_time, sizeof(string_time), format, &tm);
-        maybe_escape_markup(string_time, &outwalk);
+        outwalk += strftime(buffer, 512, format, &tm);
```
stapelberg, line 67: We can make this 4096 (which is the size of `buffer`)

cherti: Yes, having slept a night over it I also think that's actually the better approach here.

cherti: fixed.

stapelberg (approved):

stapelberg:
Thanks!

PR #433, opened 2020-10-09
Initialize fields with zero bytes in wireless module

### LINKED ISSUE #432: maybe_escape_markup introduces random characters
The function `maybe_escape_markup` in `src/output.c` (sometimes) introduces random characters into the processed string. So far I could observe this only with the wireless module and the essid-variable.

The behavior can be observed by introducing `printf("%s---%s\n", text, *buffer);` after https://github.com/i3/i3status/blob/master/src/output.c#L122.
One exemplary result can be seen here: https://p.nnev.de/9562

The behavior was verified on the text console, hinting at random characters, when using i3status as the status command of swaybar, this effectively destroys the formatting (besides other symptomes one can see broken utf-8-characters(?)), so it might not only be ascii characters being introduced here, although this test only indicates ascii-characters.

This issue could be reproduced on two Arch Linux installations with a hand-compiled i3status master. Interestingly, when using the packaged version of i3status, while sticking to the way the package was built, the issue did not occur there.

CC @stapelberg

cherti:
Previously, the fields in the wireless module were declared but not explicitly
initialized upon declaration. As nothing else would do so afterwards, this
could introduce random characters left over in the memory segment into
the fields. This was explicitly observed in the essid-field, but likely
a possibility for other fields as well. Hence, this commit adds explicit
initialization with zero bytes to all fields to ensure proper
termination of all fields.

Fixes #432

### DIFF
--- a/src/print_wireless_info.c
+++ b/src/print_wireless_info.c
@@ -554,13 +554,13 @@ void print_wireless_info(yajl_gen json_gen, char *buffer, const char *interface,
         }
     }
 
-    char string_quality[STRING_SIZE];
-    char string_signal[STRING_SIZE];
-    char string_noise[STRING_SIZE];
-    char string_essid[STRING_SIZE];
-    char string_frequency[STRING_SIZE];
-    char string_ip[STRING_SIZE];
-    char string_bitrate[STRING_SIZE];
+    char string_quality[STRING_SIZE] = {'\0'};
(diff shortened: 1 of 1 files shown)


orestisfl (approved):
Thanks!

Just make sure that you format it with clang-format. Essentially, you can't align the right-hand side.

cherti:
Done

orestisfl:
Thanks!

PR #437, opened 2020-11-17
disk_info: BSDs use f_bsize instead of f_frsize.

sepherosa:
Unbreak print_disk_info compilation on BSDs.

PR #441, opened 2020-12-02
print_disk_info: Use statvfs instead of statfs on OpenBSD.

cgohla:
This was necessary to compile on OpenBSD 6.7.

Interaction with #437 may need investigating.


stapelberg:
This is now merged as https://github.com/i3/i3status/commit/f3b9ad208601ecf24de95555ce1617fb5ecc52e1

PR #449, opened 2021-02-08
Fix detection of i3status version from git when using a worktree.

devkev:
I tried to build i3status in a [worktree](https://git-scm.com/docs/git-worktree) of my clone of the git repo, and was surprised when autoconf reported the version was "not-git".  The cause is `configure.ac` checking for a `.git` _**directory**_ in the srcdir.  However, when using worktrees, `.git` is a text file that contains info about where the actual `.git` directory can be found.  The fix is, of course, to adjust the test to allow `.git` to be a file.


orestisfl:
Closing because we moved to meson

PR #461, opened 2021-11-07
battery: set status to full when battery is not charging

leonixyz:
As reported in this [comment](https://github.com/i3/i3status/issues/304#issuecomment-962655488), when the reported battery status is:

    POWER_SUPPLY_STATUS=Not charging

The battery is actually full, therefore not charging anymore. Currently i3status shows the battery as discharging.

### DIFF
--- a/src/print_battery_info.c
+++ b/src/print_battery_info.c
@@ -185,9 +185,9 @@ static bool slurp_battery_info(battery_info_ctx_t *ctx, struct battery_info *bat
             batt_info->present_rate = abs(atoi(walk + 1));
         else if (BEGINS_WITH(last, "POWER_SUPPLY_STATUS=Charging"))
             batt_info->status = CS_CHARGING;
-        else if (BEGINS_WITH(last, "POWER_SUPPLY_STATUS=Full"))
+        else if (BEGINS_WITH(last, "POWER_SUPPLY_STATUS=Full") || BEGINS_WITH(last, "POWER_SUPPLY_STATUS=Not charging"))
             batt_info->status = CS_FULL;
-        else if (BEGINS_WITH(last, "POWER_SUPPLY_STATUS=Discharging") || BEGINS_WITH(last, "POWER_SUPPLY_STATUS=Not charging"))
+        else if (BEGINS_WITH(last, "POWER_SUPPLY_STATUS=Discharging"))
             batt_info->status = CS_DISCHARGING;
         else if (BEGINS_WITH(last, "POWER_SUPPLY_STATUS="))
             batt_info->status = CS_UNKNOWN;
--- a/testcases/023-battery-not-charging/expected_output.txt
+++ b/testcases/023-battery-not-charging/expected_output.txt
@@ -1,1 +1,1 @@
-BAT
+FULL


orestisfl (approved):

orestisfl (changes requested):
The test needs to be changed as well

leonixyz:
> The test needs to be changed as well

Fixed the test (I hope) and force-pushed.

erbth:
I think `POWER_SUPPLY_STATUS=Not charging` means that the battery is not full and a power supply (e.g. AC adapter)  is connected, but the supply does not charge the battery. Such conditions can be found in different charger/fuel gauge drivers in the Linux kernel: https://elixir.bootlin.com/linux/v5.16.13/source/drivers/power/supply/bq2515x_charger.c#L615 or https://elixir.bootlin.com/linux/v5.16.13/source/drivers/power/supply/ds2781_battery.c#L311 (note how they use the combination of power supply connected, not-full capacity and negligible charging current to determine that state). The internal enum-value `POWER_SUPPLY_STATUS_NOT_CHARGING` used in the drivers' code is translated to "Not charging" for use as context (? - not sure if this is the correct term) for uevents (and therefore /sys/.../uevent) through this array: https://elixir.bootlin.com/linux/v5.16.13/source/drivers/power/supply/power_supply_sysfs.c#L79.

Since this commit: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=44cbd2a16a07cbc02a4921853d9f17f1ffaaa97a the "Not charging" status is also reported correctly on ThinkPad laptops. (That commit made it into Debian's stable kernel with the last update.) If one has configured a charging-stop threshold on a ThinkPad, i3status previously showed UNK when that threshold was reached - now it shows BAT (with 0 watts power draw), which is in my opinion not right because the laptop is actually not powered from battery.

Therefore I would agree with https://github.com/i3/i3status/issues/304#issuecomment-962655488 that "Not charging" should not be treated as on battery, but I would also disagree with the changes in this PR, because it should not mean that the battery is full - that's what the status "Full" is for. Of course I see that the battery in https://github.com/i3/i3status/issues/304#issuecomment-962655488 is supposed to be full given the `POWER_SUPPLY_CHARGE_*` values. I would claim that the kernel driver of that charger/battery gauge of that laptop reports the "wrong" status (however one could argue that "Full" is stronger than "Not charging" and hence "Not charging" could be viewed as correct).

Referring to https://github.com/i3/i3status/issues/304#issue-352286101 I would claim that the "Not charging" status should only happen while the battery is not discharging either, hence only if a power supply is connected. However looking at the kernel code one could probably find all sorts of broken ACPI firmwares etc. out there...

Long story short I would propose adding another state to i3status's battery implementation which describes that a battery is not being charged but not full either (because it is configured to be not fully charged or because conditions prevent charging, like too high temperature). One could then map Linux's "Not charging" onto this new state. Maybe one should make this configurable to some extent (and maybe keep the behavior introduced in !305 as default) to make it easy to deal with broken firmware/driver implementations...

Of course, this relates only to Linux, I don't know about e.g. BSD.

erbth:
I've created #486 with my proposed changes; @leonixyz, @orestisfl would be great if you could have a look whether this PR might work for you/might look acceptable to you. @leonixyz with #486 it would be possible for you to specify 'FULL' for 'status_idle' in i3status.conf.

orestisfl:
Closed in favor of https://github.com/i3/i3status/pull/486

PR #462, opened 2021-11-07
Allow to build inside of a docker container

leonixyz:
In order to be able to build the project inside of a Docker container, I added a new subdirectory in the root, `docker` with a Dockerfile and instructions about how to build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,3 +42,5 @@ If you absolutely have to build from source, use:
   ninja
   sudo ninja install
 ```
+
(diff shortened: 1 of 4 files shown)


stapelberg:
Thanks for the PR! However, I think we’ll decline this one. The build dependencies are already listed in the README, and if you want to build in Docker, that gets you most of the way there (the rest is specific to your needs). Anything we merge, we need to maintain, and documenting how to build i3status in docker does not seem to address a common-enough need that it’s worth the ongoing maintenance cost.

Thanks for your understanding

PR #467, opened 2021-11-29
Add support for displaying uptime.

10maurycy10:
A patch add support for displaying uptime in the status, it is fairly minimal.


Airblader:
Thanks for the PR. However, I'm going to reject this because this is already possible using file_contents, see https://i3wm.org/docs/i3status.html#_file_contents.

10maurycy10:
> Thanks for the PR. However, I'm going to reject this because this is already possible using file_contents, see https://i3wm.org/docs/i3status.html#_file_contents.

That would print the uptime in seconds not nice units.

PR #468, opened 2021-11-29
Allow nicely displaying uptime.

10maurycy10:
Adds ``uptime`` module that displays the uptime in this format ``UP: 10 days, 23:1``.

This is not possible with ``read_file``


Airblader:
Duplicate of #467

Please don't just resubmit a PR it it doesn't get accepted. A good rule of thumb would be to create an issue first. We're very conservative with adding new modules.

PR #470, opened 2021-12-06
add sndio(7) support which is the only supported mixer backend on OpenBSD

rnagy:
This pull request adds sndio support to i3status. This is the only supported mixer backend
on OpenBSD so let's disable everything else.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -50,6 +50,10 @@ if get_option('pulseaudio')
   cdata.set('HAS_PULSEAUDIO', 1)
 endif
 
+if get_option('sndio')
+  cdata.set('HAS_SNDIO', 1)
+endif
+
 # Instead of generating config.h directly, make vcs_tag generate it so that
 # @VCS_TAG@ is replaced.
 config_h_in = configure_file(
@@ -189,6 +193,12 @@ if get_option('pulseaudio')
   pulse_dep = dependency('libpulse', method: 'pkg-config')
   i3status_deps += [pulse_dep]
   i3status_srcs += ['src/pulse.c']
+endif
+
+if get_option('sndio')
+  sndio_dep = cc.find_library('sndio')
+  i3status_deps += [sndio_dep]
+  i3status_srcs += ['src/sndio.c']
 endif
 
 host_os = host_machine.system()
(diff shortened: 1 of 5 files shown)


rnagy:
@stapelberg

meson.build
```diff
@@ -50,6 +50,10 @@ if get_option('pulseaudio')
   cdata.set('HAS_PULSEAUDIO', 1)
 endif
 
+if get_option('sndio')
```
stapelberg, line 53: Instead of making this feature optional, can you make it non-optional on OpenBSD please?

src/sndio.c
```diff
@@ -0,0 +1,195 @@
...
+static struct control *controls;
+static struct pollfd *pfds;
+
+/*
+ * new control registered or control changed
+ */
+static void ondesc(void *unused, struct sioctl_desc *d, int val) {
+    struct control *i, **pi;
+
+    if (d == NULL)
```
stapelberg, line 29: Here and elsewhere, please add braces:

```c
if (d == NULL) {
  return;
}
```

src/sndio.c
```diff
@@ -0,0 +1,195 @@
+#include <poll.h>
+#include <sndio.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include "i3status.h"
+
+struct control {
+    struct control *next;
```
stapelberg, line 9: Any reason not to use the include/queue.h macros for this list?

src/sndio.c
```diff
@@ -0,0 +1,195 @@
...
+    controls = i;
+}
+
+/*
+ * control value changed
+ */
+static void onval(void *unused, unsigned int addr, unsigned int value) {
+    struct control *c;
+
+    for (c = controls;; c = c->next) {
```
stapelberg, line 93: Why not `for (c = controls; c != NULL; c = c->next)` instead of the explicit `if (c == NULL) { return }` below?

stapelberg (changes requested):

src/sndio.c
rnagy: fixed.

src/sndio.c
rnagy: fixed

src/sndio.c
rnagy: i don't see the reason why for such simple code, but doable if you insist

stapelberg: It’s not really about complexity, but rather uniformity and code re-use. No need to have multiple linked list implementations within the same project… :)

Issue #477, opened 2022-01-07
Battery %emptytime


riazufila:
When my laptop's battery is full, the %emptytime will return as an empty string. How do I make it return a custom text or symbol?

riazufila:
![image](https://user-images.githubusercontent.com/46749569/149643619-fca8391c-1512-42b8-9042-a2372069829e.png)

This is my i3status output based on the following configurations.

```
battery all {
        format = "<span background='#458588'> %percentage %status (while != %emptytime) </span>"
        low_threshold = 30
        threshold_type = "percentage"
}
```

However, when the battery is full, `%emptytime` will return an empty string. So i3status output will appear as the following.

![image](https://user-images.githubusercontent.com/46749569/149644302-4f4d6ad2-1fdb-420a-ac10-6611cf9816d4.png)

What I want to achieve is to be able to return a custom string for `%emptytime` when battery is full.

riazufila:
I've made a PR regarding this matter.

PR #479, opened 2022-01-16
Custom battery `%emptytime` string when battery is full

riazufila:
Allow a custom string return for `%emptytime` for when battery is full. The custom string can be set from the config file in `battery` module with the option `battery_full_emptytime`.

Current behavior of `%emptytime` when battery is full is as below.

image
```
battery all {
        format = "<span background='#458588'> %percentage %status (while != %emptytime) </span>"
        low_threshold = 30
        threshold_type = "percentage"
}
```

Instead, battery module will display such as below with this PR.

image
```
battery all {
        format = "<span background='#458588'> %percentage %status (while != %emptytime) </span>"
        low_threshold = 30
        threshold_type = "percentage"
        battery_full_emptytime = "false"
}
```


riazufila:
This PR is made to entertain #477.

PR #482, opened 2022-02-02
Add new module to allow screen-brightness

erichkeane:
New module 'brightness' takes two file enpoints (like in
/sys/class/backlight) that return a numeric value, then displays this as
'brightness'.

### DIFF
--- a/i3status.c
+++ b/i3status.c
@@ -280,6 +280,17 @@ int main(int argc, char *argv[]) {
         CFG_CUSTOM_SEP_BLOCK_WIDTH_OPT,
         CFG_END()};
 
+    cfg_opt_t brightness_opts[] = {
+        CFG_STR("format", "Brightness: %brightness%", CFGF_NONE),
+        CFG_STR("brightness_path", "/sys/class/backlight/intel_backlight/brightness", CFGF_NONE),
+        CFG_STR("max_brightness_path", "/sys/class/backlight/intel_backlight/max_brightness", CFGF_NONE),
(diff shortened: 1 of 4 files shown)


erichkeane:
Hi-
First attempt at a PR for the project, but figured I'd share a feature I made a while back and have been using extensively.  Let me know how I can improve this. Thanks!

Airblader:
Thanks for your contribution. Please see https://github.com/i3/i3status/pull/184#issuecomment-264625240 for the list of duplicates of this.

erichkeane:
Huh, that is unfortunate reasoning.  I have now implemented this 3x for my personal use (and only got around to this PR this time), and find it quite useful.  Simply "looking at the screen" is insufficient to tell if the brightness is correct (and I find 'distance from 100' and 'distance from 0' extremely useful when changing brightness).  That said, I have no problem living off a fork.

Airblader:
We do have a generic i3status module to show the contents of a file now (if that is enough). In my experience, screen brightness is a much trickier subject than it seems, and I'd be somewhat worried about getting this to work across systems anyway.

erichkeane:
Sadly that module is insufficient.  The /sys/class entries of "brightness", "actual_brightness", and "max_brightness" tend to be fairly arbitrary numbers.  One of my previous laptops had a max-brightness of 3095 for some reason, and my current one is 600-something.

FWIW, this patch is somewhat generic; all it does is read the 1st numeric value from 2 different 'files', and calculates the percentage, so it is actually not particularly brightness related, though I'm not sure what else that could be used for.

PR #484, opened 2022-03-05
add tmux output format to i3status

oaken-source:
This PR is an attempt of adding tmux as a supported output format to i3status.

tmux is a terminal multiplexer that combines multiple panes and screens with a configurable status bar, where i3status can be easily integrated by specifying the following line in tmux.conf:

```
set -g status-right '#(i3status)'
```

For an unmodified i3status, this already works as expected with the output_format set to 'none'. This PR adds 'tmux' as an additional option for output_format to enable more sophisticated output, such as colors.

tmux allows color codes in the status output in the following format:

```
#[fg=#abcdef]
```

and it is possible to reset all formatting using:

```
#[default]
```

There is one curveball, and that is that tmux requires all color hex codes to be lowercase because `#F` is a reserved identifier, so the following will not work as a valid color:

```
#[fg=FF0000]
```

instead, it needs to be:

```
#[fg=ff0000]
```

This PR addresses this behavior specifically for the tmux output_format. Everything else is very straightforward.
```


primeroz:
I am really looking forward to this

PR #486, opened 2022-03-20
Add another battery status called 'idle'

erbth:
Hi everyone,

I would like to add another state to the battery-module called 'idle'. My ThinkPad has a configurable charging regulator that can be instructed to stop charging before the battery is full. Recently a Linux-kernel patch got merged s.t. this state is now correctly reported as "Not charging" by the kernel (before it was "Unknown").

However since #305 this status is mapped to "Discharging". Please have a look at https://github.com/i3/i3status/pull/461#issuecomment-1064700361, where I tried to describe why I think my proposed behavior would be better, and why I think that it would in principle also be better than the proposed change in #461. In my comment I wrote that a mechanism to customize i3status's behavior for "Not charging" might be needed, however I think the mechanism for changing the texts mapped to states through i3status.conf might be enough.

Of course, this PR covers only the foundation and Linux-specific implementation. Support for other OSs would have to be added afterwards if possible, however I cannot assist in that.

Regards,
Thomas


erbth:
I am not sure how this interacts with #304, one would have needed more information on the original problem. If you have any doubts regarding that or anything else in this PR, feel free to request me to change it.

erbth:
fyi just rebased to current main

thrasymache:
I have a system76 laptop which also lets me set a charging threshold. Mine ends up reporting UNK/Unknown along with 0.00W when it reaches this threshold. The current patch doesn't appear to make things better for me, but it also clearly doesn't make them worse.

thrasymache:
@erbth The contents of the BAT0/status pseudo-file are `Unknown`

erbth:
@thrasymache Then your driver (or hardware) does not support reporting the idle-state correctly. Good to know.

Mine (ThinkPad) shows e.g. "Not charging". Of course that depends on the kernel version, too (e.g. the acpi-driver used by mine got support only recently, currently I'm on Debian's 5.10.149-1).

src/print_battery_info.c
```diff
@@ -129,6 +130,13 @@ static void add_battery_info(struct battery_info *acc, const struct battery_info
                 acc->status = batt_info->status;
             /* else: retain FULL, since it is more specific than UNKNOWN */
             break;
+
+        case CS_IDLE:
+            if (batt_info->status != CS_UNKNOWN && batt_info->status != CS_FULL)
```
orestisfl, line 135: Not sure about this case, if the driver reports full, wouldn't it mean that the batter is now at full capacity (for some reason) and not just idle?

orestisfl (commented):
Sorry for the delayed review

src/print_battery_info.c
orestisfl: Makes sense :+1:

erbth:
thanks for merging

PR #493, opened 2022-07-21
Port memory printing on OpenBSD

lucic71:
- Split print_memory into get_memory_info which is platform dependent and print_formatted_memory which is platform independent
- Created struct print_mem_info that keeps memory counters which will be passed between get_memory_info and print_formatted_memory
- Used UVM subsystem to get memory counters on OpenBSD
- Added the requirements in README
- src/print_cpu_usage.c does not compile on OpenBSD because of the conditional definition of \_\_OpenBSD\_\_

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ On debian-based systems, the following line will install all requirements:
 apt-get install autoconf libconfuse-dev libyajl-dev libasound2-dev libiw-dev asciidoc libpulse-dev libnl-genl-3-dev meson
 ```
 
-On OpenBSD, the following line will install some of the requirements:
+On OpenBSD, the following line will install the requirements:
 ```bash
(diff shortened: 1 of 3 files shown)


lucic71:
I still need to find a way to get buffers, cached and shared memory. If you have any idea how to do that it would be much appreciated.

lucic71:
@stapelberg is this repo still active?

stapelberg:
Yes, this is still active, but I don’t really have time to answer development questions, sorry. I can try to review and merge your PR once you let me know it’s ready (it sounded like it wasn’t ready currently).

lucic71:
@stapelberg, it's ready for reviewing!

stapelberg:
> @stapelberg, it's ready for reviewing!

Thanks, this looks reasonable on first glance.

Can you format your code with clang-format to make the CI pass? Have you run the tests yet to ensure they pass, too?

Issue #497, opened 2022-08-30
Ethernet iface's with @ in ifname not properly supported


daniel-widrick:
I run my develment environments in an lxc container. Inside of my virtual environment my ethernet address in Fedora is: eth0@if24

The 'optional strip' function of print_ip_addr.c does not seem to support this type of name. 

I will submit a pull request that i believe will address the issue by looking for ':' and '@' and stripping after the first occurrence.

PR #498, opened 2022-08-30
Support Ethernet iface name's that contain @

daniel-widrick:
Pull request to address issue: #497


daniel-widrick:
I was able to get the tests to run and pass with this code on a fedora 36 vm. (the tests were upset about alsa not being a thing in my dev container).

I would view this request ready for review.

Issue #480, opened 2022-01-30
Volume config uses Pulseaudio port name instead of description


milaq:
Commit https://github.com/i3/i3status/commit/09358d26982431702c3652e8f07ebff7166a1181 introduced an odd behavior for the volume setting: Using the port name instead of the device name.
The rationale was to have "much friendlier names" which seems certainly not the case for many other systems.
Moreover, the port name can _not_ be easily renamed by the user like the device decription.

If the user find themselves in the situation of unfriendly names, they can always rename their Pulse description for sane names in i3status with the old behavior.
With the new behavior from https://github.com/i3/i3status/commit/09358d26982431702c3652e8f07ebff7166a1181 this is not possible anymore.

Additionally the rationale of the setting in i3status is (from my understanding) to show the audio _device_ not the active _port_, as the port may change while the respective volume handler stays the same.

Attached below are screenshots of both behaviors on one of my systems.
I have two sound devices with the port "Digital Output (S/PDIF)" and cannot tell them apart with the new behavior.
![screenshot_20220130_155737](https://user-images.githubusercontent.com/891620/151705016-786cd5ba-76d2-4e69-a248-82714e861e16.png)
![screenshot_20220130_155815](https://user-images.githubusercontent.com/891620/151705017-452f8ed8-abbb-496b-b837-ee5f45aee2e5.png)
![screenshot_20220130_155900](https://user-images.githubusercontent.com/891620/151705018-e93bf3b7-db9c-4ec5-a60c-75a1807164a5.png)
![screenshot_20220130_155920](https://user-images.githubusercontent.com/891620/151705020-56b0d935-f3f0-485a-9526-ded9267dffb5.png)

When reverting the commit I can easily tell them apart after renaming the device description ("AVR" and "Headphones") in Pulseaudio or Pipewire. Pavucontrol even has a gui option for that (right-click on the device).
With the new behavior I am stuck with non-identifiable names.

So here goes my humble request to revert the commit https://github.com/i3/i3status/commit/09358d26982431702c3652e8f07ebff7166a1181 or at least make the bahavor optional.

milaq:
Same for me.

Ever since I used Pipewire with both pipewire-media-session and Wireplumber, the port ID renaming is not as straightforward as changing the device description, even impossible in some configurations.
Furthermore, in pavucontrol, when assigning output/input devices, the port name does not show up, only the device description.

For a streamlined behavior, as the device description is shown in many places around the OS, I think it'd be best to revert.

[...]

PR #505, opened 2022-11-18
Revert "Use the PulseAudio port name rather than the sink name (#419)"

milaq:
Changing the port name is not as straighforward as changing the device description.
Changing the device description is easy and supported in multiple modern media backends such as pipewire-media-session and Wireplumber.

Also, the device description is unique most of the time in default deployments. The port name is often the same across different devices such as "Digital Output (S/PDIF)" or "Speakers". This leads to confusion when changing the output device via hotkeys and checking the active device via i3status in uncustomized i3status configurations.

Furthermore, in pavucontrol, when assigning output/input devices, the port name does not show up, only the device description does. The port name is not used very often around interfacing with the audio backend as opposed to the device description, which is typically used in various places around third-party apps.

This reverts commit 09358d26982431702c3652e8f07ebff7166a1181.


milaq:
See also #480

PR #512, opened 2023-05-10
battery: Add status_bat_{low,quarter,half,three_quarters}

thentenaar:
This allows for splitting status_bat by quartile, coinciding with
the FontAwesome battery icons, adding three additional thresholds to
accomodate it.

status_bat_* will only be effective if they're not set to the default
value ("BAT") in which case status_bat is used as expected, providing
backward compatibility with existing configurations.


fabpiaf:
can we add manpage updates and merge that?

Issue #496, opened 2022-08-26
Battery module incorrectly trims whitespace


njhanley:
The battery module trims leading and trailing whitespace from its final output, making it impossible to manually add padding to the format string. AFAIK this is the only module that behaves this way. It's especially problematic for users who've disabled the default separator. This behavior was introduced in b5a804d1.

Push-Buttons:
https://github.com/i3/i3status/pull/515

AdamWysokinski:
I have the same problem in the 2.15 version.

PR #515, opened 2023-09-08
[496 Battery module incorrectly trims whitespace]

Push-Buttons:
+ Added dependencies for arch-based systems in the README.md;
- Removed the trim functions in the output.c, it was only used by print_battery_info (dead code);
- Removed the forced trimming in print_battery_info; = Modifed the expected output of the 014 test. There is an space between
  the two last %parameters, the expected output must reflect this space
  character;
= print_eth_info and print_wireless_info got modifed by running
  clang-format -i **/*.[ch] (clang-format version 16.0.3) as specified
  in the CONTRIBUTING.md file;

Note:
  I believe the initial intent of the trimming was to remove spaces due
  to empty values such as in this example: %remaining having no value
  when the battery is at 100%.

  It creates a tailing space, which was in the formatting, made by the user,
  to separate the last two parameters.

  format: ... %status %remaining
  result: ... [status value][space][empty string here due to no value]

  Although, even with the trimming, an empty value right in the middle of
  a set of %parameters will still make empty spaces.

  I tested the claim other modules not trimning and it was the true. In
  other words, the trimming did not fix anything and it was causing this
  module to behave abnormally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,11 @@ On debian-based systems, the following line will install all requirements:
 apt-get install autoconf libconfuse-dev libyajl-dev libasound2-dev libiw-dev asciidoc libpulse-dev libnl-genl-3-dev meson
 ```
 
+On arch-based systems, the following line will install all requirements:
+```bash
+pacman -S base-devel alsa-lib confuse libnl libpulse yajl meson asciidoc
+```
+
 ## Upstream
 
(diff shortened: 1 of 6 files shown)


Push-Buttons:
https://github.com/i3/i3status/issues/496

PR #527, opened 2024-05-22
maybe_escape_markup: Don't mangle output if markup is disabled

jspam:
As it is, the code modifies the first char of `buffer`, increasing it by the number of bytes written.


orestisfl (approved):
Thanks

